PR #12, opened 2020-11-11
Support configuration of automountServiceAccountToken field

aaron-trout:
The issue I am trying to work around with this is that I am using the
kubernetes provider for Terraform, which is supposed to ignore the incluster
config if you provide other config, but I think there is a bug where this
does not happen.

As far as I know Atlantis itself does not need the service account for
anything so think it is fine to allow disabling this.

Obviously it would be more ideal to fix the bug in Terraform but I have a
feeling this is actually down the version of kubernetes/client-go they are
using or something. This serves as an easy workaround in the meantime.


aaron-trout:
What is your process w.r.t. bumping the chart version and making a git tag? Should I bump the version in `Chart.yaml` as part of this PR or will you do that directly once it's merged?

aaron-trout:
`✖︎ atlantis => (version: "3.12.5", path: "charts/atlantis") > Chart version not ok. Needs a version bump!`

Ah, well I guess that answers that question then 😅

aaron-trout:
Version bumped! BTW I have tried to line up the `|` in the README but blimey the table in that file in a pain to edit! Had to go fullscreen and shrink my font size to about 4 to fit the whole line on the screen to figure out if it was lined up properly!

Wondering if the more pragmatic approach is to not pad the cells in the Markdown?

lkysow (approved):

lkysow:
> Wondering if the more pragmatic approach is to not pad the cells in the Markdown?

At this point, probably that's best! Thanks for keeping the formatting though :D

Issue #7, opened 2020-10-23
Support Multiple Ingress hosts


jseiser:
I can probably submit a PR on this, if its something you would want.

Essentially, our setup creates DNS records in Route 53, and we create CNAMES to those records in CLoudflare.  So I actually need 2 ingresses, one for the `internal` entry, and one of the `external` entry.

spirosoik:
this can be closed.

PR #30, opened 2021-02-16
Add support for multiple hosts in ingress

spirosoik:
This is a breaking change so we need to bump to a major release.

Removed old:
- `ingress.host`
- `ingress.path`

and introduced:
- `ingress.hosts`
- `ingress.hosts[0].paths`

Issue: https://github.com/runatlantis/helm-charts/issues/7


lkysow:
Can we do this in a non-backwards breaking way? If `host` is not empty then use it, otherwise iterate through `hosts`.

spirosoik:
hey @lkysow sure it's possible just it adds a bit more complexity and it could be odd in the sense that there is already the `hosts`. But it's definitely feasible

spirosoik:
@lkysow have one more look. Right now we still support the old single host if it's set

spirosoik:
@lkysow any update on this?

PR #35, opened 2021-03-03
Add host aliases as an option

Trozz:
It may be required to set custom host aliases within the pods, this allows this functionality via an optional value.

Values example
```
hostAliases:
  - hostnames:
    - aaa.com
    - test.ccc.com
    ip: 10.0.0.0
  - hostnames:
    - bbb.com
    ip: 10.0.0.2
```


lkysow (dismissed):

lkysow:
Please rebase and then it looks good

Trozz:
@lkysow merged upstream, should be up to date now

Issue #38, opened 2021-03-12
No support for Azure DevOps in the Helm Chart


yohanb:
Hi!
The server supports configuration with Azure DevOps but not the Helm Chart.  Is it in the works?  When can we expect parity between the chart and the server features?
Thanks

yohanb:
PR submitted 🎉

PR #41, opened 2021-03-15
Azure DevOps integration

yohanb:
Hi,
as per issue #38, this is my contribution to add the Azure DevOps integration to the helm chart.
Cheers!


lkysow (approved):
Looks great! I think just bump up the version and we're good.

charts/atlantis/values.yaml
```diff
@@ -37,6 +37,14 @@ orgWhitelist: <replace-me>
 #   baseURL: https://bitbucket.yourorganization.com
 # (The chart will perform the base64 encoding for you for values that are stored in secrets.)
 
+# If using Azure DevOps, specify like the following:
+# azuredevops:
```
spirosoik, line 41: it's better to have a flag like `enabled: false`

spirosoik (approved):
good job!!!!

charts/atlantis/values.yaml
lkysow: yeah that's a better pattern but the other VCS stanzas don't have that so I think this is more consistent.

Eugst:
Can this one will be fixed and merged, please?

Issue #60, opened 2021-05-20
Create a chart for 0.17


courteouselk:
Current latest version of this chart refers `0.16.1` while the latest one is `0.17`.  Please create a new helm-chart version for it.

billabongrob:
Submitted a PR a few days back for it.  #56

fishpen0:
This can be closed now?

billabongrob:
> This can be closed now?

I would say so. Branch merged with changes.

PR #61, opened 2021-05-27
Bumping the default version to 17.

billabongrob:
Seems like when you try to PR a master to a master, it's failing Linting?  Seems multiple other tests are failing as well below this PR for the same reason.  Also, random swap file left in the repo. 🤷‍♂️  Trying this for #60


billabongrob:
@lkysow Let me know what you think.

PR #68, opened 2021-06-11
Allow the services LoadBalancerIP to be set

gezb:
This pull request allows the services `LoadBalancerIP` to be set via the helm chart

This is useful when deploying into a cluster where MetalLB is running and allows you to specify a fixed IP for the load balancer


charts/atlantis/values.yaml
```diff
@@ -160,6 +160,7 @@ readinessProbe:
 service:
   type: NodePort
   port: 80
+  loadBalacerIP:
```
lkysow, line 163:
```suggestion
  loadBalancerIP: null
```
Mispelled.

lkysow (changes requested):

charts/atlantis/values.yaml
gezb: Fixed

PR #85, opened 2021-09-08
fix: update apiversion for ingress. fixes #75

### LINKED ISSUE #75: Helm chart doesn't support current Ingress api
With Kubernetes 1.19 Ingress has been moved to stable networking.k8s.io/v1. Current atlantis helm chart uses old Ingress beta resource. 

An example of universal helm chart could be found in prometheus repository: https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/templates/alertmanager/ingress.yaml

dad264:
This change updates the maximum apiVersion for ingress if the Kubernetes cluster Atlantis is running on supports it.

### DIFF
--- a/charts/atlantis/templates/ingress.yaml
+++ b/charts/atlantis/templates/ingress.yaml
@@ -1,7 +1,9 @@
 {{- if .Values.ingress.enabled -}}
 {{- $fullName := include "atlantis.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
-{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" -}}
+apiVersion: networking.k8s.io/v1
+{{- else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta" -}}
 apiVersion: networking.k8s.io/v1beta1
 {{- else -}}
 apiVersion: extensions/v1beta1


dad264:
Also, this updates the method for checking capabilities since `.Capabilities.KubeVersion.GitVersion` is deprecated. See: https://github.com/helm/helm/blob/7d13cd5ca1f1eee080c180be6c3201e3a81ff6e9/pkg/chartutil/capabilities.go#L80

lkysow (dismissed):

tszym:
Hi, thanks for this patch!
It's a nice first step but the chart might still fail on a stable API because of changes in supported fields.

The chart is using `spec.rules[0].http.paths[0].backend.serviceName` and `spec.rules[0].http.paths[0].backend.servicePort` with became `spec.rules[0].http.paths[0].backend.service.name` and `spec.rules[0].http.paths[0].backend.service.port.number` or `.name`

Also `spec.rules[0].http.paths[0].pathType` became a required value and is not provided yet. This one might be heavier to change, as path values are advised to use a `*` wildcard and I don't know if one of the possible filePaths support it now.

Do you think this patch could also address these parts so the linked issue is actually fixed?

dad264:
Yes. Totally forgot about the spec change. I'll work on it and reply soon.

dad264:
FWIW, I've set this up to generate based off of what's available in the cluster. @tszym you made a note that "path values are advised to use a * wildcard" and I couldn't find that in the reference for [paths](https://kubernetes.io/docs/concepts/services-networking/ingress/#path-types) but did see it for hostnames.

tszym:
Oh right, I meant the readme of this chart currently describes ingress paths with
> Path to use in the Ingress. Should be set to `/*` if using gce-ingress in Google Cloud.

I don't know if that is still possible with the stable API as I did not find it either in the reference.

tszym:
Do you know if that would be appropriate to reference an ingress class, or provide a way to do it, in the ingress resource?

dad264:
> Oh right, I meant the readme of this chart currently describes ingress paths with
> 
> > Path to use in the Ingress. Should be set to `/*` if using gce-ingress in Google Cloud.
> 
> I don't know if that is still possible with the stable API as I did not find it either in the reference.

The documentation [here](https://cloud.google.com/kubernetes-engine/docs/concepts/ingress#multiple_backend_services) makes me think that the readme of the chart is still applicable

dad264:
> Do you know if that would be appropriate to reference an ingress class, or provide a way to do it, in the ingress resource?

Were you thinking of this? https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation

I could see creating the option to define it, but to leave it nil for now. Thoughts?

tszym:
I was thinking about adding an optional `ingressClassName` field if that's ok to you. This should be enough, I don't feel like the IngressClass resource should be part of the chart but as I understand the stable API, the field on the ingress should be useful.

dad264:
Check it out when you have time and let me know.

tszym:
Looks good to me! Maybe need to add the field to the table in the readme, or it's automatic idk.

kobejn-jb (approved):
Looks, good. I've started to work on this change myself but found your PR.

Tested on k8s 1.20.7 with traefik 1.87.1

PR #98, opened 2021-11-19
Allow configuring of basic authentication

valentindeaconu:
With the [runatlantis/atlantis v0.17.5](https://github.com/runatlantis/atlantis/releases/tag/v0.17.5) release, users are allowed to configure a basic authentication. The helm chart does not allow the `livenessProbe` and `readinessProbe` to add custom headers, so this PR adds support for both the creation of credentials and probes.

In this PR you will find:
* Updated values
* Updated README
* A secret resource holding credentials
* Modified StatefulSet resource to inject the environments required by the basic authentication configuration (`ATLANTIS_WEB_BASIC_AUTH`, `ATLANTIS_WEB_USERNAME`, `ATLANTIS_WEB_PASSWORD`)
* [Removed] Modified `livenessProbe` and `readinessProbe` to add a basic authorization header

Tested it manually, deploying atlantis in an EKS cluster.

Edit:
There's no need for `livenessProbe` and `readinessProbe` to add a basic authorization header since the bug was fixed in atlantis source code [#1896](https://github.com/runatlantis/atlantis/pull/1896), so the code was removed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ The following options are supported.  See [values.yaml](/charts/atlantis/values.
 | `initContainers`                            | Containers used to initialize context for Atlantis pods                                  | `[]`                              |
 | `extraContainers`                           | Additionnal containers to use and depends of use cases. | `[]` |
 | `hostAliases[].hostnames`                   | Hostnames for host alias entry                                  | n/a                              |
-| `hostAliases[].ip`                          | IP for host alias entry                                  | `n/a                              |
+| `hostAliases[].ip`                          | IP for host alias entry                                  | n/a                               |
 | `basicAuth.username`                        | Username for basic authentication                        | n/a                               |
 | `basicAuth.password`                        | Password for basic authentication                        | n/a                               |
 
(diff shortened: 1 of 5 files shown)


JeffreyVdb:
Looks good to me. We are also in need of this configuration. `environmentSecrets` isn't something that can help us because that would mean we have to create a secret outside of the helm installation, which means that secret will not go away automatically when we helm chart is uninstalled.

valentindeaconu:
> Looks good to me. We are also in need of this configuration. `environmentSecrets` isn't something that can help us because that would mean we have to create a secret outside of the helm installation, which means that secret will not go away automatically when we helm chart is uninstalled.

You can avoid that by using the `extraManifests` field already supported by this chart. That's the way I deployed in the first place this feature, before creating this PR.

lkysow (approved):

JeffreyVdb:
> > Looks good to me. We are also in need of this configuration. `environmentSecrets` isn't something that can help us because that would mean we have to create a secret outside of the helm installation, which means that secret will not go away automatically when we helm chart is uninstalled.
> 
> You can avoid that by using the `extraManifests` field already supported by this chart. That's the way I deployed in the first place this feature, before creating this PR.

My colleague and I both missed that one 😅 Thanks for clearing that up, that's extremely useful to know for future reference. And thanks for the changes, that'll make our configuration a lot easier as well!

PR #100, opened 2021-12-02
fix: avoid '"/key.pem": must be a relative path'

joshuasimon-taulia:
the existing githubApp logic results in 
```shell
  Warning  FailedCreate      5m6s (x18 over 16m)  statefulset-controller  create Pod atlantis-0 in StatefulSet atlantis failed error: Pod "atlantis-0" is invalid: [spec.volumes[1].secret.items[0].path: Invalid value: "/key.pem": must be a relative path, spec.containers[0].volumeMounts[1].name: Not found: "github-app-key-volume"]
```
also
`{{- if index .Values.githubApp "key.pem"}}` is kind of messy with indexing values and the helm conventions strongly suggest 

> Variable names should begin with a lowercase letter, and words should be separated with camelcase

### DIFF
--- a/charts/atlantis/templates/secret-webhook.yaml
+++ b/charts/atlantis/templates/secret-webhook.yaml
@@ -10,7 +10,7 @@ metadata:
     heritage: {{ .Release.Service }}
 data:
   {{- if .Values.githubApp }}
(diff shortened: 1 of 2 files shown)


joshuasimon-taulia:
hi @lkysow. the githubApp helm logic is currently broken. please take a look at this PR

lkysow (approved):

PR #110, opened 2022-01-03
fix: use github app config from external secret

gmontanola:
Only setting `vcsSecretName` was not enough to get Github App working, since some **StatefulSet** configuration are valid only if `githubApp.key or githubApp.secret`  have values.

for now I'm using something like this

```yaml
  githubApp:
    id: 123456
    key: fake
    secret: fake

  vcsSecretName: "github-secrets"
```

`github-secrets` is a secret created outside the chart and setting dummy values for key and secret works just fine but I'm not sure if this is intended.

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: v0.19.2
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 3.17.0
+version: 3.17.1
 keywords:
 - terraform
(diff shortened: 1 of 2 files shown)


RaphaelNeumann (approved):
LGTM

And I have same issue want to use ExternalSecret to manage githubapp auth

chenrui333 (dismissed):

chenrui333:
Thanks @gmontanola for your fix, @RaphaelNeumann for your review! Cheers!

chenrui333 (approved):

chenrui333:
released as 3.17.1, let me know if that works. Thanks!

PR #117, opened 2022-01-26
update default atlantis to 0.18.2

chenrui333:
relates to https://github.com/runatlantis/atlantis/pull/2018

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: v0.18.1
+appVersion: v0.18.2
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 3.15.1
+version: 3.15.2
 keywords:
 - terraform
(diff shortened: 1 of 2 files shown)


charts/atlantis/Chart.yaml
```diff
@@ -1,8 +1,8 @@
-apiVersion: v1
+0.18.2apiVersion: v1
```
lkysow, line 1: @chenrui333 this is malformed

charts/atlantis/values.yaml
```diff
@@ -1,4 +1,4 @@
-## -------------------------- ##
+0.18.2## -------------------------- ##
```
lkysow, line 1: same here

chenrui333: oh, i used the vscode replace. let me correct it

chenrui333: should be good now (codespace is still not ideal for such operations 😮‍💨 )

lkysow (approved):

PR #123, opened 2022-02-17
feat: support targetPort configuration in service

Tyrion85:
Currently, `targetPort` [is hardcoded in service template](https://github.com/runatlantis/helm-charts/blob/main/charts/atlantis/templates/service.yaml#L25) to value `4141` (hardcoded value for atlantis container port).

However, in certain scenarios it may be beneficial to customize `targetPort` for service - for example, when running a reverse proxy as a sidecar in front of atlantis for auth purposes.


lkysow (dismissed):

chenrui333 (approved):

chenrui333:
released as 3.17.0

PR #125, opened 2022-02-24
Switch atlantis url to https when relying on ingress

### LINKED ISSUE #124: runatlantis chart should favor https urls
The ingress that atlantis defines supports `https`, so there's really no reason for the comments it generates to include an `http` url, it's just asking for trouble.

https://github.com/runatlantis/helm-charts/blob/94926f83eb3b47eeb38c563272fe65cbdf823156/charts/atlantis/templates/statefulset.yaml#L223

jsoref:
closes #124

### DIFF
--- a/charts/atlantis/templates/statefulset.yaml
+++ b/charts/atlantis/templates/statefulset.yaml
@@ -220,7 +220,11 @@ spec:
             value: {{ .Values.atlantisUrl }}
           {{- else if .Values.ingress.enabled }}
           - name: ATLANTIS_ATLANTIS_URL
+            {{- if .Values.ingress.tls }}
+            value: https://{{ .Values.ingress.host }}
+            {{- else }}
             value: http://{{ .Values.ingress.host }}
+            {{- end }}
           {{- end }}
           {{- if .Values.github }}
           - name: ATLANTIS_GH_USER


charts/atlantis/templates/statefulset.yaml
```diff
@@ -220,7 +220,7 @@ spec:
             value: {{ .Values.atlantisUrl }}
           {{- else if .Values.ingress.enabled }}
           - name: ATLANTIS_ATLANTIS_URL
-            value: http://{{ .Values.ingress.host }}
+            value: https://{{ .Values.ingress.host }}
```
lkysow, line 223: this is a backwards incompatible change, I'm not sure it makes sense to do this. This can still be set to https if you want, correct?

jsoref: I'm not sure how you define "backwards incompatible change".

If someone is setting the `tls` flag in their ingress, their ingress is probably redirecting the `http` traffic to `https` (certainly ours does), which means that the client pretty much already has to support both `https` and following redirects. Users aren't going to run clients that don't.

Can you provide an explanation of how you'd imagine a user would configure their system to not want this behavior?

lkysow (approved):

charts/atlantis/templates/statefulset.yaml
lkysow: guarding by tls makes sense

PR #131, opened 2022-03-03
Add Common Labels

### LINKED ISSUE #122: Feature Request: Add Labels To all Resources
A high level `commonLabels` value that adds labels would be very nice to have to avoid having to append those everywhere, for people working within orgs that require tagging of k8s resources for cost/ownership. example implementation: https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/templates/_helpers.tpl#L93

philnielsen:
This allows a user to supply their own labels to all deployed kubernetes resources. These can be required for compliance reasons or informational reasons. I've tested on my own cluster but let me know if there is anything you need me to do more to make this easier to merge.

Closes #122


philnielsen:
The Lint error isn't from my change, but seems to have triggered from me changing the README. Happy to fix in this PR if it isn't too much creep but holding off for now.

lkysow (approved):

lkysow:
> The Lint error isn't from my change, but seems to have triggered from me changing the README. Happy to fix in this PR if it isn't too much creep but holding off for now.

Sorry for delay responding. If you can fix it in this PR that would be much appreciated.

philnielsen:
@lkysow no worries! Thanks for making a sweet tool! I fixed the linting error and hope you can review when you get a moment.

PR #135, opened 2022-03-17
feat: allow specifying targetPort in helm config

rayterrill:
We bundle oauth2_proxy in front of atlantis - allowing us to specify targetPort in the helm config would allow us to do this vs a hardcoded 4141.


chenrui333:
relates to #123

rayterrill:
Looks like this can be closed - #123 was merged in already which implements this.

PR #143, opened 2022-05-16
incld. configmap checksum as annotation (#74)

### LINKED ISSUE #74: How to add config/checksum from values.yaml
I am using upstream atlantis-helm chart for deploying atlantis on Kubernetes.
Created a configmap in Values.extraManifests
As I do  not  have charts on my local, I can-not add annotation to Statefulset directly.
So when I tried to add this annotation in Values.podTemplate.annotations, I am getting error.

**Annotation**
`
checksum/config: {{ include (print $.Template.BasePath "/aws-configmap.yaml") . | sha256sum }}
`
**Error** 
`
Error: failed to parse values.yaml: error converting YAML to JSON: yaml: invalid map key: map[interface {}]interface {}{"include (print $.Template.BasePath \\\"/aws-account-configmap.yaml\\\") . | sha256sum":interface {}(nil)}
`

Also if i add this annotation in single quote
`
checksum/config: '{{ include (print $.Template.BasePath "/aws-configmap.yaml") . | sha256sum }}'
`
It will create a deployment but no use as it wont restart pod after any change in configmap.

olia-dev:
Fixes #74.
Using [best practice from helm](https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments) for automatically rolling deployments.

### DIFF
--- a/charts/atlantis/templates/statefulset.yaml
+++ b/charts/atlantis/templates/statefulset.yaml
@@ -26,9 +26,11 @@ spec:
 {{- if .Values.podTemplate.labels }}
 {{ toYaml .Values.podTemplate.labels | indent 8 }}
 {{- end }}
-      {{- with .Values.podTemplate.annotations }}
       annotations:
-{{ toYaml . | indent 8 }}
+        checksum/config: {{ include (print $.Template.BasePath "/configmap-config.yaml") . | sha256sum }}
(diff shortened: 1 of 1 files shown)


kumarsen26:
Hi,
I may have missed it, but how do we give in values.yaml?

PR #155, opened 2022-06-09
feat: Allow defining priorityClassName

mubarak-j:
Add the ability to define [priorityClassName](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#pod-priority) for Atlantis StatefulSet


charts/atlantis/values.yaml
```diff
@@ -198,6 +198,7 @@ statefulSet:
     fsGroup: 1000
     runAsUser: 100
     fsGroupChangePolicy: "OnRootMismatch"
+  priorityClassName: {}
```
lkysow, line 201: shouldn't this be an empty string?

lkysow (changes requested):

charts/atlantis/values.yaml
mubarak-j: you're right. it has been updated.

Issue #93, opened 2021-10-19
Ignore ssl verify as an option?


npsoni88:
I am trying to use the helm chart to deploy atlantis and connect to internal gitlab. I am running into an error where its basically a "x509: signed by unknown authority". I don't find any option to skip sslverify in the helm chart, is it possible?

GMartinez-Sisti:
Looks like this was fixed on https://github.com/runatlantis/helm-charts/pull/163. Can we close this?

npsoniembark:
my apologies for the delay. you are correct, we can close this.

Issue #82, opened 2021-08-18
Support custom CA certs


bsecker:
There doesn't seem to be a way to use custom certs in the helm chart.

```
kubectl -n atlantis logs atlantis-0
Error: initializing server: Get "https://gitlab.[local-domain]/api/v4//version": x509: certificate signed by unknown authority
```

I have a custom root CA that signs the certs. One way it could be added would be with something like this in `values.yaml`:

```
CASecret: secret-name
```

which references a k8s secret containing the certificate.

Cheers

PR #163, opened 2022-07-08
feat: allow override of ca-certificates.crt

papanito:
Relates to #93, Implements #82

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,12 +144,14 @@ The following options are supported.  See [values.yaml](/charts/atlantis/values.
 | `commonLabels`                        | Add Common Labels to all resources | `{}` |
 | `containerSecurityContext.allowPrivilegeEscalation` | Whether to enable privilege escalation           | n/a                               |
 | `containerSecurityContext.readOnlyRootFilesystem`   | Whether the root file system should be read-only | n/a                               |
+| `customPem`   | SecretName of the custom `ca-certificates.cert` to override the `/etc/ssl/certs/ca-certificates.crt` with your custom one (self-signed certificates)<br>Secret has to be created manually and shal contain `ca-certificates.crt: PEM` | n/a                               |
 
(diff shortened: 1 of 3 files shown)


papanito:
@lkysow can you have a look at this PR?

lkysow (approved):

papanito:
Thanks a lot!

papanito:
@lkysow Could you create a new release, containing my change? Would really appreciate it

zbialik:
I could only get this to work if my Secret was generated having the `key=ca-certificates.crt`

Otherwise, I got an error:

`Error: failed to create containerd task: OCI runtime create failed: container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: rootfs_linux.go:75: mounting "/var/lib/kubelet/pods/xxxxx/volume-subpaths/additional-trust-certs/atlantis/4" to rootfs at "/etc/ssl/certs/ca-certificates.crt" caused: mount through procfd: not a directory: unknown`

I believe this is because the `kubelet` assumes `/etc/ssl/certs/ca-certificates.crt` is a **directory** if the subpath doesn't match the filename designated in the Secret key... or something like that.

If this is the case, maybe this should be documented or perhaps enforced another way? Ideally, we'd like to be able to add custom CA's and have those appended to the ones already provided out-the-box without using an `initContainer` or building a custom image.

papanito:
> If this is the case, maybe this should be documented

I can do that

> Ideally, we'd like to be able to add custom CA's and have those appended to the ones already provided out-the-box without using an initContainer or building a custom image.

Yeah this was my first thought, but if you add your certs with a configmap, you would have to run `update-ca-certificates`, which you cannot to with the user of Atlantis image.
My assumption was - as this is my setup - internal self-signed cerificates are only signed by the internal CA, so all other root ca's are irrelevant

Issue #166, opened 2022-07-14
ingress doesn't forward any path except `/`


gurpalw:
latest `4.0.5` chart.

```
ingress:
  annotations:
    alb.ingress.kubernetes.io/backend-protocol: HTTP
    alb.ingress.kubernetes.io/target-type: 'ip'
    alb.ingress.kubernetes.io/listen-ports: '[{"HTTPS": 443}]'
    alb.ingress.kubernetes.io/scheme: internet-facing
    alb.ingress.kubernetes.io/security-groups: sg-redacted, sg-redacted
    alb.ingress.kubernetes.io/manage-backend-security-group-rules: "true"
  enabled: true
  host: atlantis.dev.com #redacted real value
  ingressClassName: alb
```

Using the above values in my `values.yaml` results in only the `atlantis.dev.com/` path being resolved by the alb. `atlantis.dev.com/events` and `atlantis.dev.com/status` all fail with 404 not found.

see below image of the resulting alb listener rule in aws when using the above values.
  
![image](https://user-images.githubusercontent.com/37112979/179115732-2a14d5bb-f910-4a1e-be31-7605eda2cefa.png)

Instead i had to add the below values in addition to get it to forward all paths after the slash.

```
ingress:
  path: /*
  pathType: ImplementationSpecific
```
 
the alb ingressClassName is referring to my deployment of the aws load balancer controller chart version 1.4.1

PR #180, opened 2022-08-21
Add /* as a default ingress path

nitrocode:
## what
- Add /* as a default ingress path

## why
- This will allow access to / and other common atlantis api paths

## references
- closes https://github.com/runatlantis/helm-charts/issues/166

### DIFF
--- a/charts/atlantis/values.yaml
+++ b/charts/atlantis/values.yaml
@@ -216,7 +216,7 @@ ingress:
   annotations: {}
     # kubernetes.io/ingress.class: nginx
     # kubernetes.io/tls-acme: "true"
-  path: /
+  path: /*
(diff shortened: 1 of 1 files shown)


lkysow:
Why can't you just make this change in your own values file?

kvanzuijlen:
@lkysow This change didn't work for me with the default configuration. Also, ImplementationSpecific and Prefix are the same thing for the Nginx Ingress controller. Imo this change should therefore be unmerged and maybe handled in documentation instead.

nitrocode:
@lkysow I suppose someone could just override the values.

Apologies @kvanzuijlen for breaking your implementation. 

The default would work well for AWS load balancer controller but not for nginx. Perhaps the middle ground is to document both deployments?

PR #188, opened 2022-09-16
Make container resources optional

hazim1093:
We are seeing CPU throttling with atlantis because of the CPU limits. 
It would be nice to make the resources (requests & limits) optional, so they can be configured per use case

### DIFF
--- a/charts/atlantis/values.yaml
+++ b/charts/atlantis/values.yaml
@@ -246,13 +246,14 @@ ingress:
 # You have to create a secret `my-ca-certificates`
 # customPem: my-ca-certificates
 
-resources:
-  requests:
-    memory: 1Gi
-    cpu: 100m
-  limits:
-    memory: 1Gi
-    cpu: 100m
+resources: {}
+# resources:
+#   requests:
+#     memory: 1Gi
+#     cpu: 100m
+#   limits:
+#     memory: 1Gi
+#     cpu: 100m
 
 ## Embedded data volume & volumeMount (default working)
 volumeClaim:
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: v0.19.8
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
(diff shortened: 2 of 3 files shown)


charts/atlantis/values.yaml
```diff
@@ -246,13 +246,14 @@ ingress:
...
-    memory: 1Gi
-    cpu: 100m
+resources: {}
+# resources:
+#   requests:
+#     memory: 1Gi
+#     cpu: 100m
+#   limits:
+#     memory: 1Gi
+#     cpu: 100m
```
lkysow, line 256: This is a breaking change. Please leave it as-is and then in your own values you can set them to `{}`

lkysow (changes requested):

lkysow (commented):
Wait, actually can't you just use:

```
resources: {}
```


?

hazim1093:
Hi @lkysow, 

In our case, we just want to remove CPU limits, and provide the rest. Since having CPU limits is causing unnecessary throttling on Atlantis

lkysow:
So just set resources.limits.cpu to null. Have you tried that? You shouldn't need to change the default settings for everyone else. You should be able to override settings in your own config file.

hazim1093:
It is not possible to set it to null. Also, I believe it's not the standard practice to force certain resource requests as default values. Probably many if not all, must be having CPU throttling because of the limits.

lkysow:
```yaml
resources:
  limits:
    cpu: null
```

```
 helm template . -f test-values.yaml | grep resources -A 5                                                                                         ⎈ gke-1
          resources:
            limits:
              memory: 1Gi
            requests:
              cpu: 100m
              memory: 1Gi
--
      resources:
        requests:
          # The biggest thing Atlantis stores is the Git repo when it checks it out.
          # It deletes the repo after the pull request is merged.
          storage: 5Gi
```

It unsets it in my testing.

hazim1093:
I was trying directly in the chart files and it rendered a `cpu: null` 
Works fine with the external values.yaml, thanks 👍

lkysow:
Ahh. This has more info: https://helm.sh/docs/intro/using_helm/. See
customization.

PR #198, opened 2022-10-13
Add mergeable-bypass-apply flag

JParisR:
Hello!

This PR adds the `ATLANTIS_GH_MERGEABLE_BYPASS_APPLY` envVar in order to be able to use the mergeable mode with required apply status check when deploying atlantis with the chart.


nitrocode:
@JParisR why was this closed?

JParisR:
A little bit later I found that you can just set the variable in the extraArgs [here](https://github.com/runatlantis/helm-charts/blob/main/charts/atlantis/values.yaml#L383), so there is no need to add a new EnvVar

PR #203, opened 2022-10-28
feat: Adding .netrc support for authentication with private registries

### LINKED ISSUE #202: Add support for .netrc file creation to authenticate with http/https repositories/registries
As stated in the terraform docs - https://developer.hashicorp.com/terraform/language/modules/sources#http-urls
```
If an HTTP/HTTPS URL requires authentication credentials, use a .netrc file in your home directory to configure these.
```
This helm chart should support the creation of a .netrc file. The approach would be similar to the .gitconfig setup

nxf5025:
Signed-off-by: Nick Fisher <nxf5025@gmail.com>

Fixes #202 

This PR adds the ability for users to create a .netrc file to authenticate with http/https registries.


charts/atlantis/templates/secret-netrc.yaml
```diff
@@ -0,0 +1,10 @@
+{{- if .Values.netrc}}
```
nitrocode, line 1: Instead of using a .netrc file, couldn't you use this flag as an env var `--write-git-creds` ?

https://www.runatlantis.io/docs/server-configuration.html#write-git-creds

nxf5025: Let me look into it, but for my exact use case we need to pull the modules from Artifactory. So technically git creds don't come into play here.

charts/atlantis/templates/statefulset.yaml
```diff
@@ -399,6 +417,11 @@ spec:
             readOnly: true
             mountPath: /etc/secret-gitconfig
           {{- end }}
+          {{- if or .Values.netrc .Values.netrcSecretName }}
+          - name: netrc-volume
+            readOnly: true
+            mountPath: /etc/secret-netrc
```
nitrocode, line 423: Why not just mount it `/home/atlantis/.netrc`

nitrocode: I suppose this is following the same method as the gitconfig so it's probably fine.

charts/atlantis/templates/statefulset.yaml
```diff
@@ -128,12 +137,21 @@ spec:
...
+          {{- if or .Values.gitconfig .Values.gitconfigSecretName .Values.netrc .Values.netrcSecretName }}
           lifecycle:
             postStart:
               exec:
-                command: ["/bin/sh", "-c", "cp /etc/secret-gitconfig/gitconfig /home/atlantis/.gitconfig && chown atlantis /home/atlantis/.gitconfig"]
-          {{- end}}
+                command:
+                  - "/bin/sh"
+                  - "-c"
+                  {{- if and (or .Values.gitconfig .Values.gitconfigSecretName) (or .Values.netrc .Values.netrcSecretName) }}
```
nitrocode, line 147:
```suggestion
                  {{- if or .Values.gitconfig .Values.gitconfigSecretName .Values.netrc .Values.netrcSecretName) }}
```

charts/atlantis/templates/statefulset.yaml
```diff
@@ -128,12 +137,21 @@ spec:
...
+                  - "cp /etc/secret-gitconfig/gitconfig /home/atlantis/.gitconfig && chown atlantis /home/atlantis/.gitconfig && cp /etc/secret-netrc/netrc /home/atlantis/.netrc && chown atlantis /home/atlantis/.netrc"
+                  {{- else if or .Values.gitconfig .Values.gitconfigSecretName }}
+                  - "cp /etc/secret-gitconfig/gitconfig /home/atlantis/.gitconfig && chown atlantis /home/atlantis/.gitconfig"
+                  {{- else if or .Values.netrc .Values.netrcSecretName }}
+                  - "cp /etc/secret-netrc/netrc /home/atlantis/.netrc && chown atlantis /home/atlantis/.netrc"
+                  {{- end }}
```
nitrocode, line 153:
```suggestion
                  {{- if or .Values.gitconfig .Values.gitconfigSecretName }}
                  - "cp /etc/secret-gitconfig/gitconfig /home/atlantis/.gitconfig && chown atlantis /home/atlantis/.gitconfig"
                  {{- end }}
                  {{- if or .Values.netrc .Values.netrcSecretName }}
                  - "cp /etc/secret-netrc/netrc /home/atlantis/.netrc && chown atlantis /home/atlantis/.netrc"
                  {{- end }}
```

nxf5025: If a user sets both the gitconfig and netrc then the suggested change won't work. It will only execute the first command. For example it would be the equivalent of running the following and only the first output being executed:
```
$ /bin/sh -c "echo 'test'" "echo 'test2"
test
```

charts/atlantis/templates/statefulset.yaml
nxf5025: @nitrocode - I just tested both the gitconfig and the new netrc setup with a direct mount instead of using the lifecycle block. Seems to be working just fine. Was there a reason the gitconfig wasn't directly mounted in the first place? If not I'll push a change to directly mount instead.

nxf5025: Ran a few tests this week and didn't hit any issues so I just pushed up the direct mount

nitrocode: You can set this env var instead

https://www.runatlantis.io/docs/server-configuration.html#write-git-creds

nxf5025: That doesn't fit my use case for a few reasons:
1. It only writes out the creds for the baseURL that's set. This doesn't solve my problem of pulling modules from Artifactory.
2. In some use cases we need to use a different set of credentials so we can authenticate with vendor provided module registries. The user and token we set for our internal source control won't work for those.

charts/atlantis/templates/statefulset.yaml
nitrocode: The suggested change says that if you set gitconfig, write it. If you set netrc, write it. If you set both, it will write both.

With your logic, it says, if you use both, write both, if you use gitconfig, write only gitconfig. If you use netrc, write only netrc.

Both the logic results in the same but i thought mine was simpler. Am i wrong here?

nitrocode: cc @jamengual @chenrui333 for an extra set of eyes

charts/atlantis/values.yaml
```diff
@@ -74,6 +74,15 @@ orgAllowlist: <replace-me>
 # If managing secrets outside the chart for the gitconfig, use this variable to reference the secret name
  # gitconfigSecretName: 'mygitconfigsecret'
 
+# When referencing Terraform modules in private repositories or registries (such as Artfactory)
+# configuing a .netrc file for authentication may be required:
+# netrc: |
```
nitrocode, line 79: Please add these new values to the readme

https://github.com/runatlantis/helm-charts#customization

charts/atlantis/templates/statefulset.yaml
nxf5025: The logic is the same, but the syntax of two cp commands as separate items in the list results in only the first cp command being executed. However, I think it’s best to get rid of this block all together and just do the direct mount

nitrocode (dismissed):

nitrocode:
cc: @lkysow @jamengual please review in case I missed something

charts/atlantis/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: v0.20.1
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 4.1.2
+version: 4.2.0
```
nitrocode, line 5: Sorry, we already released 4.2.0, please update to 4.3.0

```suggestion
version: 4.3.0
```

nxf5025:
@nitrocode @lkysow @jamengual - Anything else needed to get this released?

nitrocode (approved):

Issue #154, opened 2022-06-09
using orgWhitelist instead of orgAllowlist causes error


medwards1771:
Hello runatlantis!

Love your product. Got an issue with [this commit](https://github.com/runatlantis/helm-charts/commit/89da1af324c27bbe608052a577d60957a030e9b4) however. Based on how I read that code, `orgAllowlist` is now the favored key to use, **but `orgWhitelist` should still work.** 

Problem is, `orgWhitelist` does not work. Yesterday I deployed our kubernetes-managed atlantis using the Docker image ghcr.io/runatlantis/atlantis:v0.19.4 and the runatlantis/atlantis helm chart v4.0.3. 

Then after running `atlantis plan`, I got the message: "Error: This repo is not allowlisted for Atlantis". 

After that, I changed the key `orgWhitelist` to `orgAllowlist` in our `atlantis/values.yaml` file, deployed that change, and the error resolved. 

Are you sure that `{{ toYaml (coalesce .Values.orgAllowlist .Values.orgWhitelist) }}` returns the value for `orgWhitelist` when `orgAllowlist` is not specified? 

Please note this issue is _different than_  [#152 wrong number of args for toYaml: want 1 got 3](https://github.com/runatlantis/helm-charts/issues/152). We are using chart v4.0.3, so we have already moved past that problem.

adamstrawson:
We got caught by this too, causing Atlantis to fail for all actions. 

I'm fine with the change itself, but this shouldn't have been a `patch` version release, since it's a breaking change.

We use [Renovate](https://docs.renovatebot.com/) to handle our dependency updates, and anything that's a minor or patch versions get automatically merged, as you don't really expect breaking changes in minor or below changes. 

Even just a changelog on the GitHub Releases would have been helpful here so we can see what's changed on an update, including any breaking changes as it would save investigation when something does go wrong.

nitrocode:
Is this resolved since https://github.com/runatlantis/helm-charts/pull/183 was merged ?

PR #207, opened 2022-11-05
Remove default `orgAllowlist` value

nitrocode:
## what
- Remove default allowlist value

## why
- There is a coalesce statement that will use the `orgAllowlist` instead of the deprecated `orgWhitelist`. However, since the default value of `orgWhitelist` used to be `<replace-me>` and since it was renamed, the `coalesce()` statement would resolve to `<replace-me>` which was destructive to anyone upgrading who had no switched to the non-deprecated value name.

https://github.com/runatlantis/helm-charts/blob/89da1af324c27bbe608052a577d60957a030e9b4/charts/atlantis/templates/statefulset.yaml#L221-L222

- By removing the default value for `orgAllowlist`, the `coalesce()` statement should check `orgAllowlist` and if it is empty, it should use `orgWhitelist`. If both are null, helm should fail.

## references
- Attempts to close https://github.com/runatlantis/helm-charts/issues/154

### DIFF
--- a/charts/atlantis/values.yaml
+++ b/charts/atlantis/values.yaml
@@ -6,8 +6,8 @@
 ##   if not using an ingress, set it to the external IP.
 # atlantisUrl: http://10.0.0.0
 
-# Replace this with your own repo allowlist:
-orgAllowlist: <replace-me>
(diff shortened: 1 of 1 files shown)


nitrocode:
This may not be needed since https://github.com/runatlantis/helm-charts/pull/183 was merged

Issue #201, opened 2022-10-24
Add support for not specifying a host


davidh-unmind:
Am I being dumb or have you built the ingress template in such a way that it is not possible to specify "*" any host. During testing I don't care much for the host and found this annoying.

nitrocode:
@davidh-unmind you should be able to set the `host` to anything acceptable in the `Ingress` kind spec

https://github.com/runatlantis/helm-charts/blob/a4e46785625115d3c280fc3337754c0fa5a9f670/charts/atlantis/templates/ingress.yaml#L36

Try these values

```yaml
ingress:
  host:
    name: *
```

If that doesn't work, please propose the necessary change

nitrocode:
Oh I think I understand, you'd like to completely omit the `host` key and it may not be possible at the moment. Is that right?

https://github.com/runatlantis/helm-charts/blob/a4e46785625115d3c280fc3337754c0fa5a9f670/charts/atlantis/templates/ingress.yaml#L35-L38

PR #208, opened 2022-11-05
Allow optional host to be set

nitrocode:
## what
- Allow optional host to be set
- Fix linting issue

https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-rules

## references
- Closes https://github.com/runatlantis/helm-charts/issues/201
- https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-rules

## testing

This command used to return just `rules` but now it will return the paths without the `host` if the `host` is omitted.

```sh
helm template . -s templates/ingress.yaml \
  --set "ingress.host=" \
  --set "ingress.hosts=null" | yq .spec
```

Before this change

```yaml
rules:
```

After this proposed change

```yaml
rules:
  - http:
      paths:
        - path: /*
          backend:
            serviceName: release-name-atlantis
            servicePort: 80
```

This command returns the same result before and after this change. Only difference is that this change will put the `host` in quotes which keeps it consistent with the multiple `hosts` quoting.

```sh
helm template . -s templates/ingress.yaml \
  --set "ingress.host=example.com" \
  --set "ingress.hosts=null" | yq .spec
```

```yaml
rules:
  - host: "example.com"
    http:
      paths:
        - path: /*
          backend:
            serviceName: release-name-atlantis
            servicePort: 80
```

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: v0.20.1
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 4.1.2
+version: 4.1.3
 keywords:
 - terraform
 home: https://www.runatlantis.io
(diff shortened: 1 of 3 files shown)


jamengual (approved):

nitrocode:
@davidh-unmind please take a look and ensure that this is the desired result from issue #201

PR #209, opened 2022-11-08
Added shared namespace to helm chart

samuelbailey123:
Added optional shared namespace option to help chart. 

This adds the ability to control the namespace from inside another container.

Some use cases: 
* vault token restarts
* custom bot container for logs

Simply an enhancement to the chart, currently have to edit this on deployment for my use case. I think this would be helpful to the community and it's low risk.


samuelbailey123:
Simply an enhancement to the chart, currently have to edit this on deployment for my use case. I think this would be helpful to the community and it's low risk.

charts/atlantis/templates/statefulset.yaml
```diff
@@ -42,6 +42,9 @@ spec:
       {{- end }}
       {{- end }}
       serviceAccountName: {{ template "atlantis.serviceAccountName" . }}
+      {{- if .Values.shareProcessNamespace }}
+      shareProcessNamespace: {{ .Values.shareProcessNamespace }}
+      {{- end }}
```
nitrocode, line 47: If this defaults to `false` do we even need the if statement?

samuelbailey123: Good point @nitrocode
Would you prefer having the if statement or removing the default from false? 

I think removing the if statement would be my preferred solution, but I'll do either.

nitrocode: I think removing the if statement would be fine for now since it already defaults to `false`

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/

> Share a single process namespace between all of the containers in a pod. When this is set containers will be able to view and signal processes from other containers in the same pod, and the first process in each container will not be assigned PID 1. HostPID and ShareProcessNamespace cannot both be set. Optional: Default to false.

samuelbailey123: It is now removed 👨🏻‍💻

charts/atlantis/templates/statefulset.yaml
```diff
@@ -42,6 +42,7 @@ spec:
       {{- end }}
       {{- end }}
       serviceAccountName: {{ template "atlantis.serviceAccountName" . }}
+      shareProcessNamespace: {{ .Values.shareProcessNamespace }}
```
nitrocode, line 45: Please also bump the minor version on the Chart.yaml file

samuelbailey123: done @nitrocode

charts/atlantis/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: v0.20.1
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 4.1.2
+version: 4.1.3
```
nitrocode, line 5: This is a new feature so I think we should bump the minor version, not the patch version.

cc: @jamengual thoughts?

```suggestion
version: 4.2.2
```

charts/atlantis/values.yaml
```diff
@@ -88,6 +88,9 @@ orgAllowlist: <replace-me>
 # To reference an already existing Secret object with AWS credentials
 # awsSecretName: 'mysecretwithawscreds'
 
+# option to share process namespace with atlantis container
+shareProcessNamespace: false
```
nitrocode, line 92: Please document the new value in the README

samuelbailey123: hi @nitrocode this is now documented in the README

charts/atlantis/Chart.yaml
nitrocode: cc: @samuelbailey123

charts/atlantis/templates/statefulset.yaml
nitrocode: I believe you bumped the patch version instead of the minor version.

nitrocode:
Please fix the conflict and we can merge.

samuelbailey123:
> Please fix the conflict and we can merge.

Hi this is now resolved

nitrocode (approved):

samuelbailey123:
hi @nitrocode any idea when this will be merged?

charts/atlantis/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: v0.20.1
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 4.2.1
+version: 4.3.0
```
nitrocode, line 5:
```suggestion
version: 4.4.0
```

nitrocode (approved):

PR #211, opened 2022-11-09
feat: Add Locking DB and Redis configuration

Balsir:
This functionality exposes Locking DB and Redis configuration variables as chart values.

Setting the redis.password variable creates a secret for Redis password. 
Alternatively, set redisSecretName to use an existing Kubernetes Secret for the Redis password.

PR #212, opened 2022-11-11
fix: Quote Locking DB and Redis values

Balsir:
Adds quotes to Locking DB and Redis values.
Fixes issue where providing a number or boolean value fails Helm chart installation.

Previous PR https://github.com/runatlantis/helm-charts/pull/211

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: v0.20.1
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 4.2.0
+version: 4.2.1
 keywords:
 - terraform
 home: https://www.runatlantis.io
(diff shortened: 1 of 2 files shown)


nitrocode (approved):

nitrocode:
@Balsir did you test this helm chart addition and verified that it worked correctly?

Balsir:
> @Balsir did you test this helm chart addition and verified that it worked correctly?
@nitrocode Yes, I tried it with and without `| quote`, it works correctly with `| quote`.

nitrocode:
@Balsir the redis password reference doesn't need to be quoted?

Balsir:
@nitrocode  Thanks for merging.
Good question. Passing a boolean or int64 value indeed fails the Helm chart - this is a case not just for `redis.password`, but for all secrets that are sent to the `b64enc` function.
Error from `gitlab.secret: 12345`:
```
Error: template: atlantis/templates/secret-webhook.yaml:19:119: executing "atlantis/templates/secret-webhook.yaml" at <b64enc>: wrong type for value; expected string; got int64
```
It could get patched globally in all secrets or a disclaimer could be added, but given the low likelihood of someone passing just a number in those values, I don't think it requires immediate action. What do you think?

nitrocode:
We can cross that bridge if it becomes a problem I suppose

PR #218, opened 2022-11-17
shareProcessNamespace hot patch

samuelbailey123:
Hot fix from 
https://github.com/runatlantis/helm-charts/pull/209
the value from`shareProcessNamespace` was pulling from the wrong location. This is a hot patch for it.


samuelbailey123:
Hi @nitrocode we missed the location of the process namespace. It pulled from the root of the values when it should have been under stateful set. since it was making the change in that template. Sorry for the miss.

nitrocode (approved):

PR #219, opened 2022-11-18
feat: Enable modification of AWS credentials directory path

Balsir:
Enables modification of AWS credentials directory path.
If `aws.directory` value is set:
- AWS credentials are mounted to `aws.directory` path instead of ~/.aws/
- AWS_SHARED_CREDENTIALS_FILE and AWS_CONFIG_FILE are set to their respective values

This PR is a workaround for AWS CLI writing its cache into the read-only ~/.aws/ directory, causing it to fail (https://github.com/aws/aws-cli/issues/1804).


nitrocode (approved):
This looks good to me.

Please look this over in case I missed something cc: @runatlantis/maintainers

charts/atlantis/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: v0.20.1
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 4.5.1
+version: 4.6.0
```
nitrocode, line 5: Before merging, we'll have to merge the default branch here and then update this line for the next minor version.

nitrocode (approved):

Issue #222, opened 2022-12-02
gitconfig credential.helper error: Resource busy


kpettijohn:
It seems with the changes made in #203 Atlantis now throws an error when attempting to modify the git config.

https://github.com/runatlantis/helm-charts/pull/203/files#diff-a3b675e5633e3d9602003c23ede42b91c32bc10f4aed39235b204cff10c11bbdL131

```
{"level":"error","ts":"2022-12-02T16:15:07.444Z","caller":"events/command_runner.go:169","msg":"Error running pre-workflow hooks There was an error running git config --global credential.helper store: error: could not write config file /home/atlantis/.gitconfig: Resource busy\n: exit status 4. Proceeding with plan command."}
```

Overall I believe this is due to changing the mount path to be `/home/atlantis/.gitconfig` instead of mounting the file to `/etc/secret-gitconfig` and copying the file to `/home/atlantis/.gitconfig`. This causes the `/home/atlantis/.gitconfig` file to be read only so when [Atlantis attempts to run `git config --global credential.helper store`](https://github.com/runatlantis/atlantis/blob/main/server/events/git_cred_writer.go#L55) it results in the error `Resource busy`.

It seems due to this error it causes all additional pre workflow hooks to be skipped which causes additional issues. I'm not totally sure what the best path forward is here. In our setup we include the git credential configuration as part of the `gitconfig`.

```
            gitconfig: |
              [credential]
                helper = store
              [url "https://x-access-token@github.com"]
                insteadOf = ssh://git@github.com
              [url "https://x-access-token@github.com/"]
                insteadOf = git@github.com:
```

Any ideas to fix or work around the issue would be appreciated!

nitrocode:
Thanks @kpettijohn for bringing this to our attention.

@kpettijohn @nxf5025 could you see if there is a small change to fix this?

GMartinez-Sisti:
Possible fix on https://github.com/runatlantis/helm-charts/pull/384. Please review.

bdellegrazie:
Just FYI, this isn't fixed as the volume mount is still mounted read-only.

GMartinez-Sisti:
> Just FYI, this isn't fixed as the volume mount is still mounted read-only.

Yes! We are aware of that 😓 Recent discussions were [here](https://github.com/runatlantis/atlantis/issues/1257). Re-opening !

GMartinez-Sisti:
[...]

PR #223, opened 2022-12-07
Restore gitconfig related changes from #203 (netrc support)

kpettijohn:
Changes revert gitconfig changes from #203. I believe this should fix the issues in #222.


charts/atlantis/templates/statefulset.yaml
```diff
@@ -138,6 +138,12 @@ spec:
         - name: {{ .Chart.Name }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          {{- if or .Values.gitconfig .Values.gitconfigSecretName }}
+          lifecycle:
+            postStart:
+              exec:
+                command: ["/bin/sh", "-c", "cp /etc/secret-gitconfig/gitconfig /home/atlantis/.gitconfig && chown atlantis /home/atlantis/.gitconfig"]
```
nitrocode, line 145: Doesn't the `chown` command require i.e. `user:group` e.g. `atlantis:atlantis` ?

kpettijohn: I don't believe it's required to provide the group for the chown command to work. I believe in this case the file permissions would end up being `atlantis:root`. Overall I took what was done [previously](https://github.com/runatlantis/helm-charts/pull/203/files#diff-a3b675e5633e3d9602003c23ede42b91c32bc10f4aed39235b204cff10c11bbdL135) but happy to add the group if thats preferred.

charts/atlantis/templates/statefulset.yaml
```diff
@@ -138,6 +138,12 @@ spec:
         - name: {{ .Chart.Name }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          {{- if or .Values.gitconfig .Values.gitconfigSecretName }}
```
nitrocode, line 141: You mentioned

>I believe this should fix the issues in #222.

Could you test out your version of the chart and confirm this works as you intend it to?

nitrocode: @kpettijohn

charts/atlantis/templates/statefulset.yaml
```diff
@@ -138,6 +138,12 @@ spec:
         - name: {{ .Chart.Name }}
```
nitrocode, line 138: If your intention is to revert #203 then why overwrite the file? Why not simply revert the pr ?

cc: @nxf5025 since you are the original pr author of #203

charts/atlantis/templates/statefulset.yaml
kpettijohn: I have not had the time to verify the changes at work but hopefully will after the holidays.

charts/atlantis/templates/statefulset.yaml
nxf5025: Hey sorry missed the previous notification for this. I agree with @kpettijohn on reverting just the gitconfig changes and not the .netrc as that's confirmed working properly

nitrocode (dismissed):

charts/atlantis/templates/statefulset.yaml
nitrocode: Please verify before we merge

nitrocode: @kpettijohn friendly ping

nitrocode: If we cannot test this and get this change (or similar) merged, we'll have to revert #203 @kpettijohn .

If #203 is reverted, @nxf5025 would you be able to take this over to resubmit #203 and prevent #222?

nxf5025: @nitrocode That's fine with me

Balsir: Hey, reverting to using `lifecycle.PostStart` for setting .gitconfig could clash with already merged #220

azend:
I came here to submit my own PR but it looks like this is a work in progress. For whatever reason, `gitconfigSecretName` nor an extraVolumeMount to `/home/atlantis/.gitconfig` worked for us. I believe it's because the execution environment of our Atlantis + Terragrunt config isn't setting `$HOME` (as well as a bunch of other shell defaults) for the usual Linux daemon security reasons. What does work is mounting the volume mount to the system level config path (`/etc/gitconfig`). Storing secrets in a version controlled `values.yaml` is bad practice for a number of obvious reasons. Storing secrets externally shouldn't be a second class feature. Let's do whatever it takes to get this feature functional.

nitrocode:
Hi @azend , please feel free to open another pr or fix this one. We seem to be caught in the middle of a broken pr and the inability to revert the previous pr that caused this issue.

azend:
I'm new to this project so I'm just getting up to speed on these issue tickets. It looks like the issue root cause right now is that the chart is mounting `/home/atlantis/.gitconfig` as a secret causing it to become mounted read only and with the wrong permissions. The previous fix for this was to mount the secret to an intermediate location then copy and re-permission the file to the correct location at some other point (like a pre-workflow task). This will require more testing with other users but it seems to me we can kill two birds with one stone if we just put the config directly into `/etc/gitconfig`. The config will be in a looked up path regardless of what user or context Atlantis is run from and also it frees up `~/.gitconfig` for any additional customization.

mattjamesaus:
Just following up here, this seems like a pretty hefty bug (we rolled back our chart to get this to work). Is there something i can help with here to get this fixed up?

charts/atlantis/templates/statefulset.yaml
nitrocode: @kpettijohn friendly ping

@mattjamesaus see this thread. There is still changes that need to be addressed. Please feel free to propose a new pr if you can fix these issues with the feedback

bigjew92:
Having this issue as well with the gitconfig. Any idea on if this will be merged at some point or if there is a work around?

azend:
There’s a few workarounds for this. One is the workaround mentioned above. The other is that `gosu` destroys the HOME environment variable as part of setting up the image. Setting a HOME path in your helm chart replaces the destroyed variable and reinstates intended behaviour.

bigjew92:
Thanks for the reply. I was able to solve this last night by using a similar method as this https://github.com/runatlantis/helm-charts/issues/288#issuecomment-1521087686

jamengual:
> There’s a few workarounds for this. One is the workaround mentioned above. The other is that `gosu` destroys the HOME environment variable as part of setting up the image. Setting a HOME path in your helm chart replaces the destroyed variable and reinstates intended behaviour.

We removed gosu from the image.

Issue #224, opened 2022-12-14
Wrong default value for statefulSet.shareProcessNamespace 


lacebal:
Documentation says that the default value for the `statefulSet.shareProcessNamespace` should be `false` but the key `shareProcessNamespace` default value is defined at the root level instead of `statefulSet`. 

With current definition the default value generated for `statefulSet.shareProcessNamespace` is empty.

```
helm template --values charts/atlantis/test-values.yaml charts/atlantis/. | grep shareProcessNamespace
      shareProcessNamespace: 
```

PR #225, opened 2022-12-14
fix: default shareProcessNamespace  moved from root to statefulSet

lacebal:
https://github.com/runatlantis/helm-charts/issues/224

```
helm template --values charts/atlantis/test-values.yaml charts/atlantis/. | grep shareProcessNamespace
      shareProcessNamespace: false
```


charts/atlantis/values.yaml
```diff
@@ -219,6 +216,8 @@ statefulSet:
     fsGroupChangePolicy: "OnRootMismatch"
   priorityClassName: ""
   updateStrategy: {}
+  # option to share process namespace with atlantis container
+  shareProcessNamespace: false
```
nitrocode, line 220: Could you also bump the patch version in the chart yaml?

nitrocode (approved):

PR #226, opened 2022-12-28
Require GitHub configuration only if bot user is used

Teko012:
When using GitHub Enterprise, `github.hostname` has to be set. If Atlantis uses a GitHub App for it's setup, the `github.user`, `github.token`,`github.secret` values are not needed, and it causes issues and conflicts when it's set.

To avoid this, the `github.user` value should be checked instead of just `github`, since the other two values are required for that setup.

Because `github.user` is already a required field, this should keep the chart backwards compatible, and it shouldn't be a breaking change, but it should let future chart versions use GitHub Enterprise with a GitHub App.

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: v0.21.0
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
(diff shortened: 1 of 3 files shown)


nitrocode (approved):
Looks good to me!

Teko012:
@nitrocode Great, checks have also passed, can you merge it please? :)

Issue #230, opened 2022-12-29
Automatically generate README from helm values


Teko012:
Ideally, the README should be automatically generated by the CI. 

Some of the possible options for tooling:
- https://github.com/norwoodj/helm-docs
- https://github.com/kubepack/chart-doc-gen
- https://github.com/bitnami-labs/readme-generator-for-helm
- https://github.com/rapidsai/frigate

Once the tool has been decided, the `values.yaml` can be reformatted accordingly.

Teko012:
I went through the chart to see what are the missing values in `values.yaml`, which need to be added first (I was only looking if something can be supplied as value but missing from the `values.yaml` file).

The values that are missing from the `README` currently and from the `values.yaml`:
```
basicAuthSecretName
fullnameOverride
nameOverride
```
These are in the `README`, but not mentioned in the values.yaml:
```
command
config
environment
ingress.apiVersion
orgWhitelist
service.annotations
service.loadBalancerSourceRanges
```

All of this will need to be added to the `values.yaml` file. I haven't checked if there are any that are still mentioned in the `README`, and not in the code, but I don't think that matters, and it will be fixed when it gets auto generated.

Teko012:
I think [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm) might be better for this, but [helm-docs](https://github.com/norwoodj/helm-docs) could work too, both lets you specify types.

Teko012:
We should also unify what is declared by default and what is not: https://github.com/runatlantis/helm-charts/pull/238 We will have to check if declaring it as an empty value will change some behaviour or not.

GMartinez-Sisti:
I'll take this one.

PR #238, opened 2023-01-02
Fix github default value missing

### LINKED ISSUE #236: .Values.github.user cannot be evaluated in 4.8.2
As mentioned in https://github.com/runatlantis/helm-charts/commit/02acd2515c39a33a261762d8f384a9d528752a16#r94788606 this check does not work if no `github` key is present in the `values.yaml` file.
We currently use Gitlab, so no `github` configuration is required. This leads to the following error when upgrading to 4.8.2:

```
executing "atlantis/templates/statefulset.yaml" at <.Values.github.user>: nil pointer evaluating interface {}.user
```

I think the check should either be reverted to `{{- if .Values.github }}`, an additional check on `github.user` added to the previous check, or an empty map `github: {}` added to the `values.yaml`.
I'd be willing to provide a PR, if that helps.

Teko012:
## what
Fix github default value missing

## why
It fixes #236 

## tests
n/a

## references
- closes #236
- closes https://github.com/runatlantis/helm-charts/pull/235

### DIFF
--- a/charts/atlantis/values.yaml
+++ b/charts/atlantis/values.yaml
@@ -11,6 +11,7 @@ orgAllowlist: <replace-me>
 # logLevel: "debug"
 
 # If using GitHub, specify like the following:
+github: {}
 # github:
 #   user: foo
 #   token: bar
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: v0.21.0
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 4.9.1
+version: 4.9.2
 keywords:
 - terraform
 home: https://www.runatlantis.io


charts/atlantis/values.yaml
```diff
@@ -11,6 +11,7 @@ orgAllowlist: <replace-me>
 # logLevel: "debug"
 
 # If using GitHub, specify like the following:
+github: {}
```
Jeinhaus, line 14: Would it make sense to define each provider here for consistency?

Teko012: We could, but there are some other missing values as well: https://github.com/runatlantis/helm-charts/issues/230 I think I will try to fix that in a separate PR to make it more uniform.

Teko012: (I checked in this case, it shouldn't change the behaviour, but declaring the empty values for others could make it fail, so I would rather not include it here)

nitrocode (approved):

nitrocode:
Thank you!

Teko012:
@Jeinhaus Can you test the new release?

Jeinhaus:
Will do tomorrow morning. Thank you for the fix!

Jeinhaus:
Yes, it worked for us. Thanks alot!

Teko012:
Great! Sorry for the trouble. Thanks for reporting!

PR #251, opened 2023-01-07
chore(workflows): define setup-helm action patch version

Teko012:
## what
Bump helm to the latest version

## why
The pipeline is failing, because `3.10` is not a valid version number. It should be `3.10.0` or `3.10.3`, which is the latest.

## references
[Failed run](https://github.com/runatlantis/helm-charts/actions/runs/3859256677/jobs/6578609819)
[Helm 3.10.0](https://github.com/helm/helm/releases/tag/v3.10.0)
[Helm 3.10.3](https://github.com/helm/helm/releases/tag/v3.10.3)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -22,7 +22,7 @@ jobs:
       - name: Install Helm
         uses: azure/setup-helm@v3
         with:
-          version: v3.10 # renovate: datasource=github-releases depName=helm packageName=helm/helm
+          version: v3.10.3 # renovate: datasource=github-releases depName=helm packageName=helm/helm
 
       - name: Run chart-releaser
         uses: helm/chart-releaser-action@v1


.github/workflows/release.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
       - name: Install Helm
         uses: azure/setup-helm@v3
         with:
-          version: v3.10 # renovate: datasource=github-releases depName=helm packageName=helm/helm
+          version: v3.10.3 # renovate: datasource=github-releases depName=helm packageName=helm/helm
```
nitrocode, line 25: Can we omit the version so we're always using the stable version?

https://github.com/Azure/setup-helm#example

Teko012: I guess there is a pro and con to it as there is a note there about the API rate limit, which if the runner hits, it defaults back to their hard coded version, which might not be the latest. If we control the version, it's guaranteed that it won't run it with an older version then the previous build because of it.

Teko012: @nitrocode Should I update it then?

nitrocode: I defer to the other @runatlantis/maintainers here

Teko012, line 25:
```suggestion
          version: latest
          token: ${{ secrets.GITHUB_TOKEN }}
```

Teko012: I added the other option as a suggestion.

Teko012:
@nitrocode What should we do here?

nitrocode (approved):

nitrocode:
Thank you @Teko012 !

PR #253, opened 2023-01-19
Update index.yaml

perryao:
## what

Make version 4.10.1 available in the helm repo

## references

related to #252

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -2,6 +2,26 @@ apiVersion: v1
 entries:
   atlantis:
   - apiVersion: v1
+    appVersion: v0.22.3
+    created: "2023-01-19T16:03:43.018222-05:00"
+    description: A Helm chart for Atlantis https://www.runatlantis.io
+    digest: d3b7d967955d4c793bbb739b53afb287e59075cf4d6279465a4fe2dfc194b84c
+    home: https://www.runatlantis.io
+    icon: https://www.runatlantis.io/hero.png
+    keywords:
+    - terraform
+    maintainers:
+    - name: lkysow
+    - name: jamengual
+    - name: chenrui333
+    - name: nitrocode
+    name: atlantis
+    sources:
+    - https://github.com/runatlantis/atlantis
+    urls:
+    - https://github.com/runatlantis/helm-charts/releases/download/atlantis-4.10.1.tgz
+    version: 4.10.1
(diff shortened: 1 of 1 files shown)


GenPage:
@perryao Thank you for the PR, do you mind regenerating it? I've been troubleshooting our chart-releaser and we've had some releases since. I rather you get the contribution credit.

perryao:
@GenPage happy to do so, but I'm starting to think there might be something off with the chart-releaser action, as well. Shouldn't it be updating this file as part of the release workflow? 

Instead, it seems to not be detecting any changes https://github.com/runatlantis/helm-charts/actions/runs/3969254151/jobs/6804102492#step:5:47

Looks like they're up to version 1.5.0 of that action. Maybe we could try updating to see if that helps?

perryao:
@GenPage just tested on my fork and version 1.5.0 of the chart-releaser action seems to be detecting the new releases properly https://github.com/perryao/helm-charts/actions/runs/3970538435/jobs/6806383823#step:5:45

GenPage:
@perryao Might have to do with the missing GITHUB_TOKEN secret on your fork. I believe the issue is that the index.yaml is behind after doing multiple releases. The workflow fails when it cant fast-forward, but I would expect it to. I'm still doing some investigating. We can bump the version just in case anyways

perryao:
@GenPage I regenerated it so it includes both the `atlantis-4.10.1` and `atlantis-4.10.2` releases.

nitrocode:
@perryao could you fill out the "why" section in the pr body?

Looks like this update could / should be automated with a GitHub action. Otherwise we'd have to update it upon every release, no?

GenPage:
@nitrocode This is related to the chart-releaser action. It does it for us and what was failing previously. I was hoping manually bringing index up-to-date would fix the issue on next action run but I'm not so sure after reading https://github.com/helm/chart-releaser-action/issues/60

nitrocode:
Hmm i wonder if it's working but just not updating the index.yaml in the main branch and instead using the gh-pages branch?

https://github.com/runatlantis/helm-charts/blob/gh-pages/index.yaml

GenPage:
@nitrocode That is where the index is suppose to be updated, it's not supposed to be in main branch. The index is what lists our releases to the helm binary that we host on gh-pages.

PR #255, opened 2023-01-24
chore(ci): bump chart-releaser action to v1.5

GenPage:
## what
- Chart-releaser seems to be failing to update the index.yaml in the gh-pages branch. 

## why
Hoping a combination of bumping the action and also manually bringing the index.yaml update up to date will help.

## references

https://github.com/runatlantis/helm-charts/pull/253
https://github.com/runatlantis/helm-charts/actions/runs/3969254151/attempts/1

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -27,6 +27,6 @@ jobs:
           version: v3.10.3 # renovate: datasource=github-releases depName=helm packageName=helm/helm
 
       - name: Run chart-releaser
-        uses: helm/chart-releaser-action@v1
+        uses: helm/chart-releaser-action@v1.5.0
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"


dgteixeira:
From the discussion in #253 , I'm not entirely sure that this will fix the issue.

While reading the error:
> hint: Updates were rejected because the remote contains work that you do
hint: not have locally. This is usually caused by another repository pushing
hint: to the same ref. You may want to first integrate the remote changes
hint: (e.g., 'git pull ...') before pushing again.

I don't know your default GITHUB TOKEN permissions, but you might need to add the `write` permissions as seen here in their documentation:
image

Maybe this could be the problem?

perryao:
aha! that looks pretty promising imo @dgteixeira. Nice catch.

The example for the chart-releaser-action calls this out, as well: https://github.com/helm/chart-releaser-action#example-workflow

dgteixeira:
yeah @perryao, that was where I got the info, forgot to detail my sources sorry 😅
I'm just trying to help out, because we are avid Atlantis users and we want to try out the most recent features of allowing specific commands only to defined GitHub teams (which is a very nice feature for us) :)

Any help you need just let me know!

perryao:
> yeah @perryao, that was where I got the info, forgot to detail my sources sorry 😅 I'm just trying to help out, because we are avid Atlantis users and we want to try out the most recent features of allowing specific commands only to defined GitHub teams (which is a very nice feature for us) :)
> 
> Any help you need just let me know!

oh no worries @dgteixeira. I realized after I commented that your screenshot was already showing what I linked to. That's what I get for skimming :)

GenPage:
@perryao @dgteixeira Appreciate both of your support and contributions. The GITHUB_TOKEN is configured correctly. (see: https://github.com/runatlantis/helm-charts/actions/runs/3969254151/jobs/6804102492#step:1:19)

jamengual (approved):

PR #259, opened 2023-01-31
use standard label for helm chart name and version

NitriKx:
## what

The helm documentation defines a standard label for the chart name and version: https://helm.sh/docs/chart_best_practices/labels/#standard-labels

In this PR I've added the standard label but not removed the old one to not introduce any breaking change

## why

We have monitoring tools that relies on this value to have a helm chart inventory across all our clusters. 

## tests

`helm template . -f values.yaml --debug`

In the output I've ensured that the label is not in any `matchLabels`

## references

https://helm.sh/docs/chart_best_practices/labels/#standard-labels


GenPage (approved):
Thank you for this small QoL improvement.

PR #263, opened 2023-02-28
Add support for installing a GitHub App under an organization

tampakrap:
## what

Adding a new optional environment variable `ATLANTIS_GH_ORG`, which is equivalent to the `--gh-org` argument.
Relevant documentation: https://www.runatlantis.io/docs/server-configuration.html#gh-org

## why

The installation of the GitHub App under a GitHub organization needs this extra argument / env var on the initial run.
Relevant documentation: https://www.runatlantis.io/docs/access-credentials.html#github-app

After the initial run, and after the GitHub App is installed under the desired organization, this env var can be commented out / unset again.

## tests

I have tested locally by successfully creating a new GitHub App under an organization.

## references

https://www.runatlantis.io/docs/server-configuration.html#gh-org
https://www.runatlantis.io/docs/access-credentials.html#github-app


charts/atlantis/values.yaml
```diff
@@ -8,6 +8,7 @@
 
 # Replace this with your own repo allowlist:
 orgAllowlist: <replace-me>
+#githubAppOrg: myorgname
```
GenPage, line 11: Could we move this to githubApp?

GenPage (changes requested):

charts/atlantis/values.yaml
tampakrap: Unfortunately we can't be under githubApp, because the org needs to be set up before the github app is created and set up in atlantis

jseiser:
Cant you just set this under `https://github.com/runatlantis/helm-charts/blob/main/charts/atlantis/values.yaml#L384`

tampakrap:
> Cant you just set this under `https://github.com/runatlantis/helm-charts/blob/main/charts/atlantis/values.yaml#L384`

sure, that is what I'm currently doing

charts/atlantis/values.yaml
GMartinez-Sisti: Hi @tampakrap, can you please clarify why it wouldn't work? We can still set the value on the `ATLANTIS_GH_ORG`, it's just a different name for the variable.

PR #265, opened 2023-03-01
Updating for release 0.23.0

jamengual:
## what

Sync for Atlantis 0.23.0 release

## why

new release

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: v0.22.3
+appVersion: v0.23.0
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 4.10.3
+version: 4.11.0
 keywords:
 - terraform
 home: https://www.runatlantis.io
--- a/charts/atlantis/values.yaml
+++ b/charts/atlantis/values.yaml
@@ -113,7 +113,7 @@ serviceAccountSecrets:
 image:
   repository: ghcr.io/runatlantis/atlantis
   # if not set appVersion field from Chart.yaml is used
-  tag: ""
+  tag: v0.23.0
   pullPolicy: IfNotPresent
 
 ## Optionally specify an array of imagePullSecrets.
--- a/charts/atlantis/values.yaml
+++ b/charts/atlantis/values.yaml
@@ -113,7 +113,7 @@ serviceAccountSecrets:
 image:
   repository: ghcr.io/runatlantis/atlantis
   # if not set appVersion field from Chart.yaml is used
-  tag: v0.23.0
+  tag: latest
   pullPolicy: IfNotPresent
 
 ## Optionally specify an array of imagePullSecrets.


charts/atlantis/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: v0.22.3
+appVersion: v0.23.0
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 4.10.3
+version: 4.11.0
```
nitrocode, line 5: Please also update the chart value

https://github.com/runatlantis/helm-charts/blob/a565c6a99be45d2bb9d973a791e2c3ca9fd2eaa1/charts/atlantis/values.yaml#L115

Please make it point to `latest` so it's one less thing we need to do when creating new releases. `latest` always points to the latest stable release.

charts/atlantis/values.yaml
```diff
@@ -113,7 +113,7 @@ serviceAccountSecrets:
 image:
   repository: ghcr.io/runatlantis/atlantis
   # if not set appVersion field from Chart.yaml is used
-  tag: ""
+  tag: v0.23.0
```
nitrocode, line 116:
```suggestion
  tag: latest
```

nitrocode (approved):

charts/atlantis/Chart.yaml
dschunack: Hi @nitrocode @jamengual

it's not an good idea to use the latest tag. Kubernetes will never pull the new image with the latest tag if the POD starts on the same node again and if the pullPolicy is set to "IfNotPresent". Leave it empty as before and the appVersion will used. It's not best practice to use the latest tag for Helm Chart with semantic versioning. You don't have a fixed releases with a fixed appVersion/Container Image.

Please, revert this to have a stable and fixed release with a fixed images that relies to the corresponding chart Version.

https://vsupalov.com/docker-latest-tag/

nitrocode: We can switch it to Always pull. This would allow latest and less maintenance for us.

The image tag is also overwritable, no?

For now, I put this PR in https://github.com/runatlantis/helm-charts/pull/266

steveizzle:
Hey guys, we also use your helm chart and i think its not a good idea to set a latest tag on an helm chart. We use helm chart for our lifecycle of different components and so also manage updates/rollbacks with helm charts. Also i would say its against common convention and how do you manage versioning yourself, for example if a newer version of you application needs specific changes in  a template file? Helm charts can be a pita to manage, but if you offer one, please do it the proper way...

dschunack:
@nitrocode I will fully agree on that what @steveizzle says here. If a old helm chart sets maybe some args that are maybe not available in newer atalanis versions means, the next PODs restart will fail. A newer image as expected will be used and you loose the control to have a fix/stable version running in a environment. Newer images are automatically used without knowing that and it's not possible to switch back. It's not possible to say which version is used. If the atlantis POD restarts in the night and pulled a newer version with a BUG without knowing that, means a outage of the CI/CD pipeline. 
The usage of the latest tag is no best practice and not suitable for production environments and the most common applications like Grafana, prometheus, loki, OTel and so on and on followed the semantic versioning best practices. 

It better to use the AppVersion instead of the latest tag.

nitrocode:
Ok thank you for your comments here. We'll put in a pr to revert it back to the app version

dschunack:
@nitrocode Thanks :-)

PR #266, opened 2023-03-02
fix: set pullPolicy to `Always` with `latest` tag

nitrocode:
## what

- set pullPolicy to `Always` with `latest` tag

## why

- since we're using latest to avoid having to update this repo for each new version of atlantis, we also need to bump the pull policy so it pulls the latest image when a new one is uploaded
- consistency with runatlantis/atlantis kustomize
- consistency with terraform aws modules atlantis

## references

- Previous pr #265 
- appVersion is optional https://helm.sh/docs/topics/charts/#the-chartyaml-file
- [latest usage in terraform-aws-atlantis module](https://github.com/terraform-aws-modules/terraform-aws-atlantis/blob/9d4e8ebb668a4b0637bcc0f6a1a2364229275c19/variables.tf#LL552C17-L552C23)
- feedback from @dschunack https://github.com/runatlantis/helm-charts/pull/265#discussion_r1122816742

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: v0.23.0
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 4.11.0
+version: 4.11.1
 keywords:
 - terraform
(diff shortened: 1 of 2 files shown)


GenPage (approved):

chenrui333:
make sense.

PR #280, opened 2023-04-11
refactor: update contidion

felipegfalcao:
## what

- Added a condition for `ingress.apiVersion` to accept only `networking.k8s.io/v1`, `networking.k8s.io/v1beta1`, or `extensions/v1beta1`.

## why

- This change was made to ensure that only supported API versions are accepted, preventing installation errors due to unsupported values.
- This is especially important for Kubernetes API resources, as their API versions are subject to change in new releases, and dynamic API versions are recommended.

## tests

- [ ] I have tested my changes by installing the chart with different values for `ingress.apiVersion`.

## references
- [Kubernetes API server dynamic API versioning](https://kubernetes.io/docs/reference/using-api/)


charts/atlantis/templates/ingress.yaml
```diff
@@ -1,12 +1,16 @@
...
+{{- else -}}
+  {{- $kubeVersion := .Capabilities.KubeVersion.GitVersion }}
+  {{- if semverCompare ">=1.19-0" $kubeVersion -}}
+    {{- $apiVersion = "networking.k8s.io/v1" -}}
+  {{- else if semverCompare ">=1.14-0" $kubeVersion -}}
+    {{- $apiVersion = "networking.k8s.io/v1beta1" -}}
+  {{- else -}}
+    {{- $apiVersion = "extensions/v1beta1" -}}
+  {{- end }}
+{{- end }}
```
felipegfalcao, line 13: This condition checks if the value of `ingress.apiVersion` is set to `networking.k8s.io/v1`, `networking.k8s.io/v1beta1`, or `extensions/v1beta1`. If the value is not set to any of these supported API versions, it sets the variable `$apiVersion` to the appropriate value based on the Kubernetes version. If the Kubernetes version is `>=1.19-0`, it sets `$apiVersion` to `networking.k8s.io/v1`. If the version is `>=1.14-0` but less than `1.19-0`, it sets `$apiVersion` to `networking.k8s.io/v1beta1`. If the version is less than `1.14-0`, it sets `$apiVersion` to `extensions/v1beta1`.

GenPage (approved):
Thank you for this. Make sure to rebase and bump the chart version please

GenPage (approved):

PR #294, opened 2023-05-22
Bump default app version to 0.24.2

jseiser:
## what

Bump App Version

## why

Default install should be latest stable release.

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -1,9 +1,9 @@
 apiVersion: v1
 # renovate datasource=docker depName=ghcr.io/runatlantis/atlantis
-appVersion: v0.23.5
+appVersion: v0.24.0
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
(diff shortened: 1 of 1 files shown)


jseiser:
just waiting on the Image to show up so i can test.

nitrocode:
We may need to bump it one more time as there is still an outstanding issue related to policy.

We really need to fix the renovatebot here too to automatically bump this https://github.com/runatlantis/helm-charts/issues/295

nitrocode:
No, let's keep this open. We may need to re-adjust this if we release 0.24.2 this week.

jseiser:
@nitrocode 

I was able to test this out and it seems correct.

jseiser:
@nitrocode 

Is this image going to be used, or will they be releasing a 0.24.2 and we should just wait?

nitrocode:
There are still some issues. I'd wait.

See https://github.com/runatlantis/atlantis/pull/3461

jseiser:
@nitrocode 

bumped to 0.24.2

PR #299, opened 2023-06-05
Remove random prefix for test pod

mustdiechik:
Also added annotations for test pod manifest.

## what
Remove random prefix for test pod name.
Also added annotations for test pod manifest.

## why
It is bad practice to user random prefix in name.

- this leads to problems when using kustomize because the searching resource is implemented via name search.
- if it necessary, random value can be passed via annotation.
- in some cases generated templates need to be without diff.
## tests

local, by compare templates
with empty and filled .Values.test.annotations

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 appVersion: v0.24.3
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 4.13.1
+version: 4.13.2
 keywords:
   - terraform
 home: https://www.runatlantis.io
--- a/charts/atlantis/templates/tests/test-atlantis-pod.yaml
+++ b/charts/atlantis/templates/tests/test-atlantis-pod.yaml
@@ -2,9 +2,12 @@
 apiVersion: v1
 kind: Pod
 metadata:
-  name: "{{ .Release.Name }}-ui-test-{{ randAlphaNum 5 | lower }}"
+  name: "{{ .Release.Name }}-ui-test"
   annotations:
     helm.sh/hook: test
+    {{- with .Values.test.annotations }}
+    {{- . | toYaml | nindent 4 }}
+    {{- end }}
 spec:
   containers:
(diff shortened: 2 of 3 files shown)


GMartinez-Sisti (approved):
Tested this locally and works fine. I was wondering if having a static name for a pod would prevent scheduling due to name conflict but it's not a problem for the `helm test` command.

GenPage:
I am unfamiliar with helm test, so I will defer here.

GenPage (approved):

mustdiechik:
what we are waiting for?

GMartinez-Sisti (approved):

GMartinez-Sisti:
Approved, but I'm not a maintainer so we still need another one.

jamengual:
can you bump the version for the chart?

mustdiechik:
@jamengual done

jamengual (dismissed):

charts/atlantis/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 # renovate datasource=docker depName=ghcr.io/runatlantis/atlantis
-appVersion: v0.24.3
+appVersion: v0.24.4
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
 version: 4.13.1
```
jamengual, line 6: you need to change this too

charts/atlantis/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 # renovate datasource=docker depName=ghcr.io/runatlantis/atlantis
-appVersion: v0.24.3
+appVersion: v0.24.4
```
jamengual, line 3:
```suggestion
appVersion: v0.24.3
```

charts/atlantis/Chart.yaml
jamengual:
```suggestion
version: 4.13.2
```

jamengual (approved):

jamengual:
thanks @mustdiechik for the contribution

PR #314, opened 2023-07-24
Add the hideUnchangedPlanComments parameter

antham:
## what

Add a parameter to be able to set the `ATLANTIS_HIDE_UNCHANGED_PLAN_COMMENTS` environment variable that allows to remove no-changes plan comments from the pull request

## why

We already have `ATLANTIS_HIDE_PREV_PLAN_COMMENTS` available so it makes sense to have both

## tests

Tested locally by generating the template


jamengual:
please bump the Chart minor version.
Thanks for the contribution @antham

antham:
done @jamengual

charts/atlantis/Chart.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: v1
 appVersion: v0.24.4
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 4.14.1
+version: 4.15.0
```
jamengual, line 6: lets do `4.14.2`

antham: Ah so the patch version

antham: done

jamengual (approved):

PR #321, opened 2023-09-14
commenting out default servicemonitor auth to avoid checkov CKV_SECRET_6

rwd5213:
## what

Commenting out the default password and username setting for service monitor authentication to avoid running into the checkov SECRET_6 compliance check.

## why

Since its disabled by default any way there should be no harm in commenting out the actual setting of the password and prevents people from deploying with a default password if they do enable the auth check for the service monitor. Commented out and left for reference of users that want to set it.

## tests

Running checkov against charts with it commented out does not make the error occur.

### DIFF
--- a/charts/atlantis/values.yaml
+++ b/charts/atlantis/values.yaml
@@ -510,8 +510,8 @@ servicemonitor:
       enabled: false
       # name: atlantis-env
       # keys:
-        # username: USERNAME
-        # password: ATLANTIS_WEB_PASSWORD
+      #   username: USERNAME
+      #   password: ATLANTIS_WEB_PASSWORD
 
 # Enable this if you're using Google Managed Prometheus
 podMonitor:
--- a/charts/atlantis/values.yaml
+++ b/charts/atlantis/values.yaml
@@ -508,10 +508,10 @@ servicemonitor:
     externalSecret:
       # authentication based on an external secret
       enabled: false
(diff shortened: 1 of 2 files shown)


charts/atlantis/values.yaml
```diff
@@ -508,10 +508,10 @@ servicemonitor:
...
       # authentication based on an external secret
       enabled: false
-      name: atlantis-env
-      keys:
-        username: USERNAME
-        password: ATLANTIS_WEB_PASSWORD
+      # name: atlantis-env
+      # keys:
+        # username: USERNAME
+        # password: ATLANTIS_WEB_PASSWORD
```
GMartinez-Sisti, line 514: nit: keep the comment blocks aligned, similar to existing ones:
```suggestion
      # name: atlantis-env
      # keys:
      #   username: USERNAME
      #   password: ATLANTIS_WEB_PASSWORD
```

GMartinez-Sisti:
Hi @rwd5213, even thought there's no actual change, it's a good practice to bump the chart. Bumping the patch version would be enough.

rwd5213:
@GMartinez-Sisti Fixed both, thanks

GMartinez-Sisti (approved):
LGTM. I approve although I’m not a maintainer. I just help around 🙂

jamengual (approved):

PR #323, opened 2023-10-17
Update values.yaml metrics

Austin37AF:
Metrics were indented one too many spaces. Fixed indentation so it is not under repos:. It should be same level as repos and not under

## what

Changed indentation on metrics section under repos. Indentation was too many spaces causing metrics to be under repos instead of at the same level in the server side config. 

## why

This caused it to not send metrics properly as it was not formatted correctly in the yaml config. 

## tests

I have tested my changes by running a local environment and passing metrics along to Grafana to make sure they are properly working. Performed Atlantis tests to make sure all metrics were being sent via statsd.


jamengual (approved):

jamengual:
@Austin37AF can you bump the chart version please?, and Thanks for the contribution

PR #332, opened 2023-11-13
Add loadBalancerSourceRanges to schema

sviatlo:
## what

Add loadBalancerSourceRanges to schema

## why

The loadBalancerSourceRanges is an existing param in the chart, but passing it to the chart after the schema was added results in `- service: Additional property loadBalancerSourceRanges is not allowed` error.

## tests

Applied with passing the value and without passing the value.

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 appVersion: v0.26.0
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 4.17.0
+version: 4.17.1
 keywords:
   - terraform
 home: https://www.runatlantis.io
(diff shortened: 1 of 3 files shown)


roytev:
@sviatlo Thanks! i missed it can you please also add it to the values so it will be noticed?

jamengual:
@sviatlo can bump the char version to 4.17.2? thanks.

sviatlo:
@roytev should I set it to null or empty array is good enough?

roytev:
Hi @sviatlo Thanks it looks good!

roytev (approved):

sviatlo:
@roytev @jamengual apologies for bugging you, but could someone please approve a CI run?

jamengual (approved):

jamengual:
@sviatlo Thanks for the contribution!

PR #337, opened 2023-11-16
Fix values.schema.json -  containerSecurityContext is an instance of io.k8s.api.core.v1.SecurityContext

milesarmstrong:
## what

Update the JSON schema to reflect that `containerSecurityContext` in `values.yaml` can be an instance of [`io.k8s.api.core.v1.SecurityContext`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#securitycontext-v1-core)

## why

To allow the full range of Security Context settings, for instance, we use `capabilities`.

## tests

Checked with https://www.jsonschemavalidator.net/

## references

N/A

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 appVersion: v0.26.0
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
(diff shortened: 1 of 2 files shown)


jamengual:
@roytev

roytev:
LGTM @jamengual

jamengual (approved):

PR #359, opened 2024-03-01
fix: Changed runAsUser uid

### LINKED ISSUE #305: Default uuid wrong if using debian image
So with the introduction of debian based image the default uuid is now wrong :
 
https://github.com/runatlantis/helm-charts/blob/3aa2c78167587ba1db03c52b21d62a098c94596d/charts/atlantis/values.yaml#L232

uuid 100 on debian is _apt vs uuid 100 on alpine being atlantis.

To be able to do things normally debian needs uuid changed to 1000.

Not sure what would be the best way to make the chart automatically guess which it should use tbh - Ive seen some projects try to guess based on image name but this seems very finicky. 

I guess we can just add a comment about this gotcha to the chart and also add an entry in the actual atlantis documentation.

kvanzuijlen:
## what
This changes the default runAsUser uid to 1000 instead of 100

## why
Together with https://github.com/runatlantis/atlantis/pull/4304 this will fix #305. This should only be merged when https://github.com/runatlantis/atlantis/pull/4304 has been released.

## references
Closes #305

### DIFF
--- a/charts/atlantis/values.yaml
+++ b/charts/atlantis/values.yaml
@@ -260,7 +260,7 @@ statefulSet:
   labels: {}
   securityContext:
     fsGroup: 1000
-    runAsUser: 100
+    runAsUser: 1000
     fsGroupChangePolicy: "OnRootMismatch"
   priorityClassName: ""
   updateStrategy: {}


kvanzuijlen:
With the latest changes to https://github.com/runatlantis/atlantis/pull/4304 this change is no longer needed

charts/atlantis/values.yaml
```diff
@@ -260,7 +260,7 @@ statefulSet:
   labels: {}
   securityContext:
     fsGroup: 1000
```
jamengual, line 262: We need to discuss if this is the right thing to do.
We made changes to Alpine and Debian to match the UID recently and we need to keep backwards compatibility
this could affect users that are not enforcing the default.

@GenPage @GMartinez-Sisti @nitrocode

jamengual (commented):
we need to

charts/atlantis/values.yaml
GMartinez-Sisti: This is going to be closed. There’s another PR fixes the uid to be 100 on both -> https://github.com/runatlantis/atlantis/pull/4304. Please let us know what you think.

nitrocode: Prior to merging #4304, is this fix #359 even needed when the values can be overridden for `runAsUser`?

kvanzuijlen: If runatlantis/atlantis#4304 gets merged in its current state this won't be necessary anymore no.

PR #304, opened 2023-06-21
Move atlantis-data volume to a separate PVC

BrunoTarijon:
## what

Remove the atlantis-data volume creation from the sts and add a separate PVC.

## why

We can not change the storage size without deleting the deployed sts.

## tests

I have tested the logic using the helm template function to generate the manifests with the volumeClaim interface and the deprecated .Values.dataStorage.

PR #356, opened 2024-02-27
feat(annotations): add global annotations

GMartinez-Sisti:
## what

- Allow adding global annotations to created resources
- Linted some files to make reading easier

## why

Requested feature.

## tests

Templates were generated against the main branch and this branch and compared to ensure there were no extra changes

```sh
→ diff --color ci-init-config-values-main.yaml ci-init-config-values-new.yaml
```
```diff
113a114
>
127,129c128,130
<         checksum/config: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
<         checksum/repo-config: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
<         checksum/init-config: da4123f44ff227abfe55354cd7872ec561edf6bc58885cca89bea77d6f533a21
---
>         checksum/config: 01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b
>         checksum/repo-config: 01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b
>         checksum/init-config: e035a1806c7275d6e396d5c7d9b2b0bd51252a161e68ef3e652ee0fc5e6576d9
```

```sh
→ diff --color ci-values-main.yaml ci-values-new.yaml
```
``` diff
77a78
>
91,92c92,93
<         checksum/config: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
<         checksum/repo-config: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
---
>         checksum/config: 01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b
>         checksum/repo-config: 01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b
```

Then I tested specifying an annotation and a secret:

```sh
→ helm template atlantis ./charts/atlantis/ --set api.secret=aaa --set extraAnnotations.aaa=bbb | diff --color old.yaml 
```
```diff
-12a13,30
>   annotations:
>     aaa: bbb
> ---
> # Source: atlantis/templates/secret-api.yaml
> apiVersion: v1
> kind: Secret
> metadata:
>   name: atlantis-api
>   labels:
>     app: atlantis
>     chart: atlantis-4.21.1
>     helm.sh/chart: atlantis-4.21.1
>     release: atlantis
>     heritage: Helm
[...]

PR #360, opened 2024-03-06
fix pvc labels indent

tetsuya28:
## what

Fix PVC `labels` indent.
Ref: https://github.com/runatlantis/helm-charts/pull/304

## why

I got the following error with `v4.22.0`.
```
Error from server (BadRequest): error when creating "STDIN": PersistentVolumeClaim in version "v1" cannot be handled as a PersistentVolumeClaim: strict decoding error: unknown field "labels"
```


GMartinez-Sisti:
This was included on https://github.com/runatlantis/helm-charts/pull/356. Just merged. Thanks!

PR #366, opened 2024-03-11
fix: fix values.schema.json for extraArgs

### LINKED ISSUE #365: extraArgs.0: Invalid type. Expected: object, given: string
Looks like there is an issue with extraArgs:

```yaml
extraArgs:
  - "--enable-regexp-cmd" # https://www.runatlantis.io/docs/server-configuration.html#flags
  - "--write-git-creds" # https://www.runatlantis.io/docs/server-configuration.html#write-git-creds
```

results into:

```
  Error: Failed to render chart: exit status 1: Error: UPGRADE FAILED: values don't meet the specifications of the schema(s) in the following chart(s):
  atlantis:
  - extraArgs.0: Invalid type. Expected: object, given: string
  - extraArgs.1: Invalid type. Expected: object, given: string
```

kvanzuijlen:
## what
Fixes #365. This changes the array item type back to string for extraArgs, which seems to have been erroneously changed in #356 

## why
Even if this wasn't erroneously, it should've been a breaking change.

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 appVersion: v0.27.0
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 4.23.3
+version: 4.23.4
 keywords:
   - terraform
 home: https://www.runatlantis.io
--- a/charts/atlantis/values.schema.json
+++ b/charts/atlantis/values.schema.json
@@ -1055,7 +1055,7 @@
       "type": "array",
       "description": "These annotations will be added to all the resources",
       "items": {
-        "type": "object"
+        "type": "string"
       },
       "examples": {
         "team": "example"


MarkDPierce:
pfft im over here troubleshooting and tracking down the problem and you ninja a PR for the change <3

GMartinez-Sisti:
Hi @kvanzuijlen, this was indeed by mistake, thank you for the fix 🙏. We can add a CI test file to create these annotations and make sure this doesn’t get reverted. If you can add it to the PR I appreciate it, otherwise we’ll do it later.

kvanzuijlen:
Hi @GMartinez-Sisti,

I was already planning on adding some CI tests (not only for this one) at a later moment, but I won't be available for the next 2 weeks.

GMartinez-Sisti:
> Hi @GMartinez-Sisti,
> 
> I was already planning on adding some CI tests (not only for this one) at a later moment, but I won't be available for the next 2 weeks.

Whenever you can it's perfect 😄. Thank you!

GMartinez-Sisti (approved):

PR #385, opened 2024-05-07
Set aws file secrets as individual files

### LINKED ISSUE #380: aws cli cache directory unwritable
The default aws cli cache directory (/home/atlantis/.aws/cli) is unwritable as the aws secret is mounted as as the entire /home/atlantis/.aws directory.

When dealing with lots of profiles, etc. it is nice to have the aws cli tool available in atlantis (we build our own image w/ additional toolsets installed).  But the helm chart's existing method for managing the aws secret doesnt allow the aws cli to work out of the box.  

I will submit a PR shortly to iterate through the aws objects and mount them individually as files within the .aws directory instead.

ryan-dyer-sp:
## what

creates config and credentials as files within .aws instead of mounting entire secret as .aws. 

## why

This allows for other directories (.aws/cli, .aws/sso) which need to be created/used by the CLI itself for caching to exist.

## tests

tested via various configuration of helm values to receive desired result.   Running in own environment the case where aws.config is specified.  Does assume that config/credentials is never specified along with awsSecretName.

## references

closes #380

### DIFF
--- a/charts/atlantis/templates/statefulset.yaml
+++ b/charts/atlantis/templates/statefulset.yaml
@@ -565,10 +565,22 @@ spec:
             mountPath: /home/atlantis/.netrc
             subPath: netrc
           {{- end }}
-          {{- if or .Values.aws.credentials .Values.aws.config .Values.awsSecretName }}
+          {{- if .Values.awsSecretName }}
           - name: aws-volume
             readOnly: true
             mountPath: {{ .Values.aws.directory | default "/home/atlantis/.aws" }}
+          {{- else }}
+          {{- range $filename, $_ := .Values.aws }}
+          {{- if has $filename (list "credentials" "config") }}
+          - name: aws-volume
+            readOnly: true
+            mountPath: {{ $.Values.aws.directory | default "/home/atlantis/.aws" }}/{{ $filename }}
+            subPath: {{ $filename }}
+          {{- else if has $filename (list "awsSecretName") }}
+          - name: aws-volume
+            readOnly: true
+            mountPath: {{ $.Values.aws.directory | default "/home/atlantis/.aws" }}
+          {{- end }}
           {{- end }}
           {{- if .Values.tlsSecretName }}
           - name: tls
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 appVersion: v0.30.0
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 5.7.0
+version: 5.7.1
 keywords:
   - terraform
 home: https://www.runatlantis.io
--- a/charts/atlantis/README.md
+++ b/charts/atlantis/README.md
@@ -79,7 +79,7 @@ extraManifests:
 | atlantisDataDirectory | string | `"/atlantis-data"` | Path to the data directory for the volumeMount. |
 | atlantisUrl | string | `""` | An option to override the atlantis url, if not using an ingress, set it to the external IP. Check values.yaml for examples. |
 | aws | object | `{}` | To specify AWS credentials to be mapped to ~/.aws or to aws.directory. Check values.yaml for examples. |
-| awsSecretName | string | `""` | To reference an already existing Secret object with AWS credentials |
+| awsSecretName | string | `""` | To reference an already existing Secret object with AWS credentials. This has priority over the aws.config and aws.credential fields. |
 | azuredevops | object | `{}` | If using Azure DevOps, please enter your values as follows. The chart will perform the base64 encoding for you for values that are stored in secrets. Check values.yaml for examples. |
 | basicAuth | object | `{"password":"","username":""}` | Optionally specify an username and a password for basic authentication. |
 | basicAuthSecretName | string | `""` | If managing secrets outside the chart for the Basic Auth secret, use this variable to reference the secret name. |
--- a/charts/atlantis/values.yaml
+++ b/charts/atlantis/values.yaml
@@ -145,7 +145,7 @@ aws: {}
 #     source_profile = default
 #   directory: "/home/atlantis/.aws"
 
-# -- To reference an already existing Secret object with AWS credentials
+# -- To reference an already existing Secret object with AWS credentials. This has priority over the aws.config and aws.credential fields.
 awsSecretName: ""
 
 # -- To keep backwards compatibility only.


GMartinez-Sisti:
Thanks @ryan-dyer-sp! Can you please provide an example on `values.yaml` and also bump the chart minor version, I would consider this a feature.

ryan-dyer-sp:
> Thanks @ryan-dyer-sp! Can you please provide an example on `values.yaml`
IDK what you're asking for here.  https://github.com/runatlantis/helm-charts/blob/main/charts/atlantis/values.yaml#L128-L139 isnt changing.  Its just mapping the individual elements within the secret to filenames instead of mounting the entire secret as a volume.
One behavior that does change with this is that secrets mounted with subPaths do not auto update (changes to the secret do not get updated within the container).  So you do have to restart the atlantis pod in this case.  Perhaps I should add the standard checksum annotation/label in order to force a redeploy when the config is updated?
Will update chart.yaml.

Thanks

charts/atlantis/templates/statefulset.yaml
```diff
@@ -522,10 +522,17 @@ spec:
...
           {{- end }}
-          {{- if or .Values.aws.credentials .Values.aws.config .Values.awsSecretName }}
+          {{- range $filename, $_ := .Values.aws }}
+          {{- if has $filename (list "credentials" "config") }}
           - name: aws-volume
             readOnly: true
-            mountPath: {{ .Values.aws.directory | default "/home/atlantis/.aws" }}
+            mountPath: {{ $.Values.aws.directory | default "/home/atlantis/.aws" }}/{{ $filename }}
+            subPath: {{ $filename }}
+          {{- else if has $filename (list "awsSecretName") }}
```
GMartinez-Sisti, line 531: The key `awsSecretName` is not within `.Values.aws`, is something expected to change or it's a mistake?

GMartinez-Sisti:
Hi!

> IDK what you're asking for here. https://github.com/runatlantis/helm-charts/blob/main/charts/atlantis/values.yaml#L128-L139 isnt changing. Its just mapping the individual elements within the secret to filenames instead of mounting the entire secret as a volume. 

Please check [my comment](https://github.com/runatlantis/helm-charts/pull/385#discussion_r1593570508), I'm trying to understand if the values are supposed to change.

> One behavior that does change with this is that secrets mounted with subPaths do not auto update (changes to the secret do not get updated within the container). So you do have to restart the atlantis pod in this case. Perhaps I should add the standard checksum annotation/label in order to force a redeploy when the config is updated? Will update chart.yaml.

This is a good idea 👍

charts/atlantis/templates/statefulset.yaml
ryan-dyer-sp: yeah just a mistake.  didnt notice it wasnt part of the aws block.

charts/atlantis/templates/statefulset.yaml
```diff
@@ -522,7 +525,15 @@ spec:
...
-          {{- if or .Values.aws.credentials .Values.aws.config .Values.awsSecretName }}
+          {{- range $filename, $_ := .Values.aws }}
+          {{- if has $filename (list "credentials" "config") }}
+          - name: aws-volume
+            readOnly: true
+            mountPath: {{ $.Values.aws.directory | default "/home/atlantis/.aws" }}/{{ $filename }}
+            subPath: {{ $filename }}
+          {{- end }}
+          {{- end }}
+          {{- if .Values.awsSecretName }}
```
GMartinez-Sisti, line 536: I would just run this if the previous is empty, otherwise it will append both and will likely fail.
```suggestion
          {{- else if .Values.awsSecretName }}
```

ryan-dyer-sp: This is a bit tricky due to the if statement being outside the range block.  I can add another if at the top, but I'm ultimately not sure what the intent is when specifying awsSecretName.  Is that intended to be mutually exclusive with the use of aws.config?  As right now if I specify both(in my own local copy), the volumeMounts get created based on the .aws.X filenames, but the volume references awsSecretName and the secret doesnt get created with that name.

GMartinez-Sisti: Yeah, the logic is not right (lint failed). An easy way is to give priority to the existing secret since it has to be created externally in before, so requires extra planning:

```tpl
{{- if .Values.awsSecretName }}
- name: aws-volume
readOnly: true
mountPath: {{ .Values.aws.directory | default "/home/atlantis/.aws" }}
{{- else }}
{{- range $filename, $_ := .Values.aws }}
{{- if has $filename (list "credentials" "config") }}
- name: aws-volume
readOnly: true
mountPath: {{ $.Values.aws.directory | default "/home/atlantis/.aws" }}/{{ $filename }}
subPath: {{ $filename }}
{{- end }}
{{- end }}
{{- end }}
```

If you decide to use this logic then a comment on values.yaml saying it has priority should be enough. This is also compatible with current implementation. WDYT?

ryan-dyer-sp:
@GMartinez-Sisti Anything else needed?

GMartinez-Sisti:
> @GMartinez-Sisti Anything else needed?

I'll look into this soon! 😃

charts/atlantis/templates/statefulset.yaml
```diff
@@ -541,10 +544,19 @@ spec:
             mountPath: /home/atlantis/.netrc
             subPath: netrc
           {{- end }}
-          {{- if or .Values.aws.credentials .Values.aws.config .Values.awsSecretName }}
+          {{- if .Values.awsSecretName }}
           - name: aws-volume
             readOnly: true
             mountPath: {{ .Values.aws.directory | default "/home/atlantis/.aws" }}
```
GMartinez-Sisti, line 550: Please add this to the `values.yaml` example so it's not hidden.

GMartinez-Sisti (commented):
Almost ready to go! Please see the comment below and also bump the minor version for the chart and it will be good to merge. Sorry for taking so long to review!

charts/atlantis/templates/statefulset.yaml
ryan-dyer-sp: I'm not sure what you're asking for here.   Just to indicate in the values.yaml that config and credentials now mount as individual files instead of a directory?

GMartinez-Sisti: I was saying to add `.Values.aws.directory` to the default values because I thought it was a new config, but looking further I see that it is not. Sorry for that.

Please bump the chart minor version so we can merge this!

ryan-dyer-sp:
/remove stale

Chart version bumped

ryan-dyer-sp:
@GMartinez-Sisti re updated.  Anything else?

GMartinez-Sisti:
> @GMartinez-Sisti re updated. Anything else?

Thanks @ryan-dyer-sp! What do you think about adding a unit test to ensure it works as expected and no one breaks it by mistake?

Issue #386, opened 2024-05-07
unable to set enabled params to the helm chart 


KyriosGN0:
Hi, in my org we set all of our helm chart with enabled params to control deployment in argocd 

how ever in atlantis values schema file there is not enabled param and we can't set additional properties. 

we would like to either have a enabled param or just be able to use additional properties

GMartinez-Sisti:
https://github.com/runatlantis/helm-charts/pull/387#issuecomment-2106269871

PR #387, opened 2024-05-07
set additionalProperties to true

KyriosGN0:
## what

allow setting additional properties in the chart 

## why

we would like the ability to add additional properties to the chart 

## references
closes [#386](https://github.com/runatlantis/helm-charts/issues/386)

### DIFF
--- a/charts/atlantis/values.schema.json
+++ b/charts/atlantis/values.schema.json
@@ -2,7 +2,7 @@
   "$schema": "https://json-schema.org/draft/2019-09/schema",
   "description": "Default values for atlantis. Declare variables to be passed into your templates.",
   "type": "object",
-  "additionalProperties": false,
+  "additionalProperties": true,
   "properties": {
     "global": {
       "description": "Global values applied to all charts",


KyriosGN0:
hi @GMartinez-Sisti, can i have some quick review on my PR? (if you would like a different approach, i can just add a enabled param in the schema file and have it be not required )

GMartinez-Sisti:
> hi @GMartinez-Sisti, can i have some quick review on my PR? (if you would like a different approach, i can just add a enabled param in the schema file and have it be not required )

Hi @KyriosGN0, sorry for the late response! I was thinking about this PR this week and I feel that allowing extra values would invalidate having a schema file, the goal is exactly to prevent having users sending wrong configurations to the chart. Adding an enabled key doesn't make sense since it wouldn't do anything.

Looks like you have a similar setup as most places I've been, having enable and versions keys is common to control deploying an app, but we didn't add at the helm chart values level, we configured the values inside another key exactly to prevent situations like this:

Values file:

```yaml
atlantis:
  enabled: true
  version: 1.0.0
  helm:
    foo: bar
```

Argocd App:

```yaml
{{- if .Values.atlantis.enabled }}
---
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: atlantis
spec:
  [...]
  source:
    chart: atlantis
    repoURL: https://runatlantis.github.io/helm-charts
    targetRevision: {{ .Values.atlantis.version }}
    helm:
      values: |
        {{- toYaml .Values.atlantis.helm | nindent 8 }}
{{- end }}
```

This way we don't have extra fields sent to the chart values.

While writing this I thought of another easy option, we can add an explicit `extra` config key to allow users to set  whatever they want, I think that having an explicit key for this doesn't invalidate the fact that we are strictly validating the other keys, since it's intentionally being used for that and not a possible mistake or typo.

Please let us know what you think of the options 😄

KyriosGN0:
Thanks @GMartinez-Sisti for the review, we use a appset to create the apps so that's why we put a want to use atlantis.enabled to decide if to deploy the app at the destination cluster. 
so i think i the `extra` key is a the way to go, i'll open modify the PR in the coming days

KyriosGN0:
after i took a look around the schema file i found i can use `global` instead of adding a new one
so closing this PR, Thanks!

PR #393, opened 2024-06-12
✨ expose init-shared-path readOnly var

vmdude:
## what

By default `init-shared-path` is read-only mounted without the possibility to mount it read-write. This PR expose the `readOnly` setting of the `volumeMount` to be able to manage it through values file.

## why

We used to build a custom `atlantis` image to be able to embed additional tools required by custom workflows defined server-side (like `aws-cli` or `asdf` to handle `terraform`, `terragrunt`, `opentofu`, `kubectl` and others). Although it works just fine, we had to maintain our custom image.
Since `initConfig` object, we have been able to move the instruction from our custom `Dockerfile` to the `initConfig.script` value.
The main issue is that the `initConfig.sharedDir` is mounted read-write on the init container, but is mounted read-only on the main container.
As we're using `asdf` is all our repositories with sometimes different versions for each tool, the `atlantis` image is installing these different versions of `terraform`/`opentofu`/`...` during the `pre_workflow_hooks` thanks to a `.tool-versions` in each repository. But `asdf` require to download and install each version where it was installed (ie in the `sharedDir`), thus require read-write `volumeMount`.

Today we're using `kustomize` to patch it so that it's mounted correctly, but we're hoping this could be handled directly through the chart itself:

```yaml
- op: replace
  path: /spec/template/spec/containers/0/volumeMounts/3/readOnly
  value: false
```

### DIFF
--- a/charts/atlantis/values.schema.json
+++ b/charts/atlantis/values.schema.json
@@ -1012,7 +1012,7 @@
         },
         "sharedDirReadOnly": {
           "type": "boolean",
-          "description": "Let you mount SharedDir in ReadWrite or ReadOnly in the main container"
+          "description": "Sets permissions level for the SharedDir in the main container to ReadWrite when true or ReadWrite when false"
         },
         "workDir": {
           "type": "string",
--- a/charts/atlantis/values.schema.json
+++ b/charts/atlantis/values.schema.json
@@ -1010,6 +1010,10 @@
           "type": "string",
           "description": "sharedDir is set as env var INIT_SHARED_DIR"
         },
+        "sharedDirReadOnly": {
+          "type": "boolean",
+          "description": "Let you mount SharedDir in ReadWrite or ReadOnly in the main container"
+        },
         "workDir": {
           "type": "string",
           "description": "Starting directory for the script"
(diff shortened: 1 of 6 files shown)


charts/atlantis/values.schema.json
```diff
@@ -1010,6 +1010,10 @@
           "type": "string",
           "description": "sharedDir is set as env var INIT_SHARED_DIR"
         },
+        "sharedDirReadOnly": {
+          "type": "boolean",
+          "description": "Let you mount SharedDir in ReadWrite or ReadOnly in the main container"
```
GMartinez-Sisti, line 1015: nit: 
```suggestion
          "description": "Sets permissions level for the SharedDir in the main container to ReadWrite when true or ReadWrite when false"
```

GMartinez-Sisti (changes requested):
Thanks @vmdude! Also please bump minor since this is a new feature and run the helm-docs update (instructions on the [README](https://github.com/runatlantis/helm-charts/blob/main/charts/atlantis/README.md#update-documentation) file).

vmdude:
Thanks for your feedback, I just pushed the required modification

GMartinez-Sisti (approved):
Thanks for the contribution 🙏

PR #401, opened 2024-07-02
Better handling of PATH env var with initConfig

vmdude:
## what

We got a use-case where we need to defined the `PATH` environment variable through the `.Values.environment` [node](https://github.com/runatlantis/helm-charts/blob/e984db5a37cee1b1eecff8bb43702105fb8227a5/charts/atlantis/values.yaml#L475) for the statefulset, and we had some issues due to duplicate entries.
When `.Values.initConfig` is enabled, the chart actually add a [`PATH` environment variable](https://github.com/runatlantis/helm-charts/blob/e984db5a37cee1b1eecff8bb43702105fb8227a5/charts/atlantis/templates/statefulset.yaml#L488).
The issue is due to how Kubernetes handle duplicate envvars definition, when a referenced key is present in multiple resources, the value associated with the last source will override all previous values.
As `.Values.initConfig` variable is defined later in the manifest, it will always supersede previously definition, thus our `.Values.environment` node was bypassed.

This PR will add the following logic:
- if `.Values.initConfig.enabled` and no `PATH` in `.Values.environment`: add a `PATH` envvar like before
- if `.Values.initConfig.enabled` and `PATH` in `.Values.environment`: do not add a `PATH` envvar as it should use the one defined in `.Values.environment`

The rest of the logic is untouched.

## why

We wanted to be able to define more specifically how the `PATH` environment variable is presented to the statefulset avoiding any duplicate in the process.

## tests

- [X] I have tested my changes by created several values file for all possible combination regarding `PATH` and `initConfig` parameters

## references

- More information about precedence and duplicate env var management [[so](https://stackoverflow.com/questions/66288565/duplicated-env-variable-names-in-pod-definition-what-is-the-precedence-rule-to)]

### DIFF
--- a/charts/atlantis/templates/statefulset.yaml
+++ b/charts/atlantis/templates/statefulset.yaml
@@ -252,8 +252,13 @@ spec:
           {{- end }}
           {{- end }}
           {{- range $key, $value := .Values.environment }}
+          {{- if and ($.Values.initConfig.enabled) ($.Values.initConfig.sharedDir) (eq $key "PATH") }}
+          - name: PATH
+            value: {{ printf "%s:%s" $value $.Values.initConfig.sharedDir | quote}}
+          {{- else }}
           - name: {{ $key }}
             value: {{ $value | quote }}
+          {{- end }}
           {{- end }}
           {{- range .Values.environmentSecrets }}
           - name: {{ .name }}
@@ -485,8 +490,10 @@ spec:
             value: {{ .Values.aws.directory }}/config
           {{- end }}
           {{- if .Values.initConfig.enabled }}
+          {{- if not .Values.environment.PATH }}
           - name: PATH
             value: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:{{ .Values.initConfig.sharedDir }}
+          {{- end }}
           - name: INIT_SHARED_DIR
             value: {{ .Values.initConfig.sharedDir }}
           {{- end }}
--- a/charts/atlantis/templates/statefulset.yaml
+++ b/charts/atlantis/templates/statefulset.yaml
@@ -251,14 +251,11 @@ spec:
           {{- toYaml . | nindent 10 }}
           {{- end }}
           {{- end }}
-          {{- range $key, $value := .Values.environment }}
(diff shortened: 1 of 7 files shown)


charts/atlantis/templates/statefulset.yaml
```diff
@@ -485,8 +490,10 @@ spec:
             value: {{ .Values.aws.directory }}/config
           {{- end }}
           {{- if .Values.initConfig.enabled }}
+          {{- if not .Values.environment.PATH }}
           - name: PATH
             value: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:{{ .Values.initConfig.sharedDir }}
```
GMartinez-Sisti, line 495: The initial paths will be lost if the user specifies a PATH: `/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:`. I'm afraid this might be hidden from the chart user. I would say we can either append them (using a named template) or add a comment for the user to know it needs to specify them, when `environment.PATH` is added in either case.

I would prefer the first since usually there is no harm in adding them if you can just prefix with your own. wdyt?

vmdude: Indeed, I would presume user setting a `environment.path` value to include default as well, but that's not awesome I agree.
Agreed with appending paths, no harm there. So if I'm understand correctly we could:
- defined a new value like `extraPath` (or something similar)
- create a new helper that will combine default system path + `initConfig.sharedDir` (if defined) + `extraPath` (if defined)
- explicitly create the environment variable `PATH` with the helper value
This way the `PATH` envvar will always be unique (no duplicate mess) AND will contain at least system required `PATH` + any additional ones ?

GMartinez-Sisti: 👍 This works. To maintain backwards compatibility I would set the new `PATH` before `{{- range $key, $value := .Values.environment }}` so we can ensure that users with a `PATH` already set and with `initConfig` disabled won't be affected as it will override the new logic.

Also please:

vmdude: Ok, I've just pushed new logic, If you could double check I'll be grateful 👍

GMartinez-Sisti (approved):
Spot on 🚀

PR #404, opened 2024-07-05
feat(chart): allow Github App installation Id

bdellegrazie:
## Github App Installation Id

* Add support for specifying this directly in the chart

## why

Supplying the `installationId` directly as part of the Github App configuration is simpler, requires less requests and works when there are multiple installations of the same Github App, as is common for Enterprise setups.

## tests

- [X] I have tested my changes by deploying Atlantis in an environment that requires an installation ID, previously using the environment variable directly.

## references

Compliments:
https://github.com/runatlantis/atlantis/pull/4650


GMartinez-Sisti:
Hi @bdellegrazie, thanks for the contribution! Can you also please:

- bump the minor field from [Chart.yaml](https://github.com/runatlantis/helm-charts/blob/main/charts/atlantis/Chart.yaml) since this is a new feature
- update the schema for [githubApp](https://github.com/runatlantis/helm-charts/blob/main/charts/atlantis/values.schema.json#L83) to include the new field
- update docs by running `make docs` from the git root folder

Thank you 🙏

bdellegrazie:
Hi @GMartinez-Sisti,

Updated as requested - thanks

GMartinez-Sisti (approved):
Thanks!

PR #411, opened 2024-08-02
Allow setting externalTrafficPolicy on service

aleksy-zalenski:
## what

This PR allows setting the externalTrafficPolicy on service used by Atlantis. 

## why

I have noticed that some webhook requests flowing through a cloud load balancer, that has externalTrafficPolicy set to Cluster can fail with 504 error. After changing the value of externalTrafficPolicy to Local, the issue was resolved, as we have removed the extra hops between the LB and Atlantis pod. 

## tests

Tested locally + run the unit tests provided.

## references

https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip


GMartinez-Sisti (approved):
Thanks @aleksy-zalenski!

PR #414, opened 2024-08-22
fix(chart): Allow setting portName on service

HanCheo:
## what

Allow setting portName field on service

## why

When using istio. there is a rule that sets the name of the service port. For this reason, we enable the portname setting.

image

## tests
local test

## references

https://github.com/istio/istio/issues/11509


GMartinez-Sisti:
Hi @HanCheo, thanks for the fix! Please check the CI, docs are not matching.

HanCheo:
HI @GMartinez-Sisti, i fix it!

GMartinez-Sisti (approved):
Thanks @HanCheo!

PR #416, opened 2024-08-28
[WIP] Gitea Config

julienym:
## what

Gitea support for the chart
PR open to seak comments, need more works

### DIFF
--- a/charts/atlantis/templates/statefulset.yaml
+++ b/charts/atlantis/templates/statefulset.yaml
@@ -387,6 +387,26 @@ spec:
             value: "/var/github-app/key.pem"
           {{- end }}
           {{- end }}
+          {{- if .Values.gitea.user }}
+          - name: ATLANTIS_GITEA_USER
+            value: {{ required "gitea.user is required if gitea configuration is specified." .Values.gitea.user }}
+          - name: ATLANTIS_GITEA_TOKEN
+            valueFrom:
+              secretKeyRef:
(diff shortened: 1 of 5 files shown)


charts/atlantis/templates/statefulset.yaml
```diff
@@ -387,6 +387,26 @@ spec:
...
           {{- end }}
+          {{- if .Values.gitea.user }}
+          - name: ATLANTIS_GITEA_USER
+            value: {{ required "gitea.user is required if gitea configuration is specified." .Values.gitea.user }}
+          - name: ATLANTIS_GITEA_TOKEN
+            valueFrom:
+              secretKeyRef:
+                name: {{ template "atlantis.vcsSecretName" . }}
+                key: gitea_token
+          {{- if .Values.gitea.baseURL }}
```
julienym, line 398: should we aim for gitea.com if not provided?

GMartinez-Sisti: I'd say, let's use the same logic as the others and leave it empty when not provided:

- https://github.com/runatlantis/helm-charts/blob/ad4e15511afe9ee70162ee13487eec98f9976539/charts/atlantis/values.yaml#L57-L62
- https://github.com/runatlantis/helm-charts/blob/ad4e15511afe9ee70162ee13487eec98f9976539/charts/atlantis/values.yaml#L71-L76

PR #423, opened 2024-09-16
chore(chart): bump initConfig script example versions

cgarcia-l:
## what
- Updated the versions in the `initConfig` script for the examples of `terragrunt` and `terragrunt-atlantis-config`.
- Bumped the `terragrunt` version from `v0.47.0` to `v0.67.5` and the `terragrunt-atlantis-config` version from `1.16.0` to `1.18.0`.
- Adjusted the download commands and SHA256 checks for the new versions of the tools.

## why
- Keeping the versions of `terragrunt` and `terragrunt-atlantis-config` up to date ensures better security, performance, and compatibility.
- The new versions include bug fixes and improvements, which enhance the functionality of the `initConfig` example.
- It provides up-to-date examples for users who want to configure their Atlantis pods with these tools.

## tests
- [ ] Verified that the downloaded binaries match the provided SHA256 sums.
- [ ] Tested the script to ensure it correctly installs and configures `terragrunt` and `terragrunt-atlantis-config` in the init container.

## references
- N/A

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 appVersion: v0.29.0
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 5.5.0
+version: 5.5.1
 keywords:
   - terraform
 home: https://www.runatlantis.io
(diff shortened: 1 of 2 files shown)


GMartinez-Sisti (approved):
Thanks @cgarcia-l. We can see from the job that the checksums matched.

We did had an issue with the first job run, but it was a timing issue with bats and atlantis being ready. If it happens again we'll have to adjust bats to wait a little bit more.

PR #425, opened 2024-09-20
fix(templates): Fix and boolean on line 44 of statefulset.yaml

### LINKED ISSUE #405: Help to use version 5.4.0 of helm chart
Hi, Everyone!

How are you?

I'm testing upgrade of my atlantis chart version of the 4.13.0 to 5.4.0, I adjusted all needed parameters but when I execute the helm
I'm receiving the following error:

```
Error: template: atlantis/templates/statefulset.yaml:44:47: executing "atlantis/templates/statefulset.yaml" at <.Values.dnsPolicy>: dnsPolicy is not a method but has arguments
```

image

I commented the parameter dnsPolicy inte the values file, but error persists.

Anyone could be a a sugestion how  solve this problemas, please?

I grateful any tip.

tmessinis:
## what

Helm templating syntax requires that the boolean keyword goes before the two values that are being compared.
Example:
```
{{if and value1 value2}}
```
instead of
```
{{if value1 and value2}}
```

## why

This fixes the issue with respect to the following error when upgrading the helm chart:
```
Error: UPGRADE FAILED: template: atlantis/templates/statefulset.yaml:44:47: executing "atlantis/templates/statefulset.yaml" at <.Values.dnsPolicy>: dnsPolicy is not a method but has arguments
```

## tests

This fix has been tested locally and the manifests render out properly.

## references

closes #405

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 appVersion: v0.29.0
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 5.5.1
+version: 5.5.2
 keywords:
   - terraform
 home: https://www.runatlantis.io
--- a/charts/atlantis/README.md
+++ b/charts/atlantis/README.md
@@ -142,7 +142,7 @@ extraManifests:
 | initConfig.script | string | Check values.yaml. | Script to run on the init container. |
 | initConfig.sharedDir | string | `"/plugins"` | SharedDir is set as env var INIT_SHARED_DIR. |
 | initConfig.sharedDirReadOnly | bool | `true` |  |
-| initConfig.sizeLimit | string | `"100Mi"` | Size for the shared volume. |
+| initConfig.sizeLimit | string | `"300Mi"` | Size for the shared volume. |
 | initConfig.workDir | string | `"/tmp"` |  |
 | initContainers | list | `[]` | Optionally specify init containers manifests to be added to the Atlantis pod. Check values.yaml for examples. |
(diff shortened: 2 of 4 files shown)


GMartinez-Sisti:
Hi @tmessinis, thanks for the fix! Unfortunately it appears there is an issue with a CI job. I can’t tell if it’s due to this change, although highly unlikely, but we need to fix it before merging this. Even if it’s on another PR.

tmessinis:
Hey @GMartinez-Sisti, if I'm reading this output correctly it looks like the issue has to do with my version bump. Should I revert that change?

https://github.com/runatlantis/helm-charts/actions/runs/10964250068/job/30453099061?pr=425#step:11:397

tmessinis:
Hey @GMartinez-Sisti, based on some local testing that I did, the size of the init-shared-path emptyDir was too small to hold both the terragrunt and terragrunt-atlantis-config binaries downloaded by the init container and the atlantis pod was getting evicted. This had the side effect of making the helm test command fail, since there was no pod for the service to forward the request to and for [curl to fail](https://github.com/runatlantis/helm-charts/actions/runs/10964250068/job/30453099061?pr=425#step:11:384) because of it.

GMartinez-Sisti:
> Hey @GMartinez-Sisti, based on some local testing that I did, the size of the init-shared-path emptyDir was too small to hold both the terragrunt and terragrunt-atlantis-config binaries downloaded by the init container and the atlantis pod was getting evicted. This had the side effect of making the helm test command fail, since there was no pod for the service to forward the request to and for [curl to fail](https://github.com/runatlantis/helm-charts/actions/runs/10964250068/job/30453099061?pr=425#step:11:384) because of it.

Awesome finding! Thanks for the effort. Please check the CI log as now we just need to sync the docs with the new change.

tmessinis:
Hey @GMartinez-Sisti, the readme has been updated. Is there something else I need to update as well?

GMartinez-Sisti (approved):

PR #434, opened 2024-10-31
Redis Secret Password Key

JohnLahr:
## what

- Added `redisSecretPasswordKey` value
- Kept previous `password` value as default for backward-compatibility

## why

This allows users to override the default `password` key used when specifying an existing `Secret` for Redis. This makes for a more seamless integration with popular third-party Redis Helm charts.

For example, the Bitnami Redis chart creates a secret called `<release>-redis` with a key called `redis-password` whose value is the password generated for that Redis distribution. Allowing users to specify an arbitrary key name means they can seamlessly integrate that Redis chart with this Atlantis chart and use the same secret.

## tests

I have tested this successfully in an Amazon EKS cluster.

## references

N/A

### DIFF
--- a/charts/atlantis/values.yaml
+++ b/charts/atlantis/values.yaml
@@ -732,7 +732,7 @@ redis: {}
 redisSecretName: ""
 
 # -- Key within the existing Redis secret that contains the password value.
-redisSecretPasswordKey: "password"
+redisSecretPasswordKey: password
 
 # -- Set lifecycle hooks.
 # https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/.
--- a/charts/atlantis/values.yaml
+++ b/charts/atlantis/values.yaml
@@ -731,6 +731,9 @@ redis: {}
 # -- When managing secrets outside the chart for the Redis secret, use this variable to reference the secret name.
 redisSecretName: ""
 
+# -- Key within the existing Redis secret that contains the password value.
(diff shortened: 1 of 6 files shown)


charts/atlantis/values.yaml
```diff
@@ -731,6 +731,9 @@ redis: {}
 # -- When managing secrets outside the chart for the Redis secret, use this variable to reference the secret name.
 redisSecretName: ""
 
+# -- Key within the existing Redis secret that contains the password value.
+redisSecretPasswordKey: "password"
```
GMartinez-Sisti, line 735: Nit: to keep it similar to other keys, we only quote empty values:
```suggestion
redisSecretPasswordKey: password
```

GMartinez-Sisti:
Hi @JohnLahr, thank you for the contribution! Would you mind adding a couple of unit test cases to test this?

GMartinez-Sisti:
> @GMartinez-Sisti Added unit tests. Not as familiar with writing tests for charts, but I think I had enough context to work with. Let me know if those aren't sufficient!

Thanks @JohnLahr! They look great.

Now there’s just one small update requirement for the docs. Please check the job that failed.

GMartinez-Sisti (approved):

GMartinez-Sisti:
> Apologies. That should be fixed now 👍

No need for apologies! Thank you again for the contribution 😁

Issue #330, opened 2023-11-10
Oauth2 proxy support for atlantis ingress


roytev:
Hi!
in order to access Atlantis ui i wanted SSO with github and limit the access to specific team members in our organization today is not possible OutOfTheBox, what i've done is deploying https://oauth2-proxy.github.io/oauth2-proxy/
what im thinking is to introduce built in support for that inside the helm chart as a sidecar or additional deployment

What do you think?

jamengual:
+1 to that, a lot of people have to do that by themselves so this could be a cool addition

bschaatsbergen:
Awesome idea!

jseiser:
I dont think this should be bundled in, as its not a core part of atlantis.  Configuring OATH support, is already available OOTB, and we do it today.

```
ingress:
  enabled: true
  ingressClassName: nginx-external
  annotations:
    nginx.ingress.kubernetes.io/service-upstream: "true"
    cert-manager.io/cluster-issuer: cert-manager-r53-prod
    nginx.ingress.kubernetes.io/auth-signin: "https://vouch-ops-${env}.xxx.network/login?url=$scheme://$http_host$request_uri&vouch-failcount=$auth_resp_failcount&X-Vouch-Token=$auth_resp_jwt&error=$auth_resp_err"
    nginx.ingress.kubernetes.io/auth-url: https://vouch-ops-${env}.xxx.network/validate
    nginx.ingress.kubernetes.io/auth-response-headers: X-Vouch-User
    nginx.ingress.kubernetes.io/auth-snippet: |
      auth_request_set $auth_resp_jwt $upstream_http_x_vouch_jwt;
      auth_request_set $auth_resp_err $upstream_http_x_vouch_err;
      auth_request_set $auth_resp_failcount $upstream_http_x_vouch_failcount;
  host: atlantis-ops-dev.xxx.network
  path: /
  tls:
    - secretName: atlantis-prod-tls
      hosts:
        - atlantis-ops-dev.xxx.network
webhook_ingress:
  enabled: true
  ingressClassName: nginx-internal
  annotations:
    nginx.ingress.kubernetes.io/service-upstream: "true"
    cert-manager.io/cluster-issuer: cert-manager-r53-prod
  host: atlantis-webhook-ops-dev.xxx.network
  path: /events
  tls:
    - secretName: atlantis-webhook-dev-tls
      hosts:
        - atlantis-webhook-ops-dev.xxx.network
```

[...]

PR #440, opened 2024-12-11
feat(oauth2-proxy): add sidecar oauth2-proxy

DrummyFloyd:
## what

oauth2-proxy sidecar

## why

Many people ask for something similar

## tests

- [x] Unit Test Only atm
- [ ] will try to test asap on my cluster (need to make a bit of work on my infra for that)

## references
Closes: #330


GMartinez-Sisti (changes requested):
Hi @DrummyFloyd, appreciate the time you took to create this, but as [this comment](https://github.com/runatlantis/helm-charts/issues/330#issuecomment-2526409920) states:

> I think adding the wrapper chart example to the docs and example folder
will be a best approach.

We prefer to not include components that are not part of the chart since we don't want to maintain things we don't control.

Can you update to include that instead?

PR #444, opened 2024-12-20
feat(atlantis): add containerPort custom

jullianow:
## what
This change seeks to provide the possibility of defining a port for the container different from the one specified in the service.
## why

This change aims to maintain compatibility with scenarios in which the Atlantis administrator wants to deploy a sidecar (using the extraContainers property) in which it will receive the traffic coming from the Service. This is a case, for example, when we want to have an oauth2-proxy responding to requests and acting as a proxy for Atlantis. With this change, this will still be possible for those who need it. By default, the behavior continues to be that the Service's targetPort is the same value for the Atlantis ContainerPort.


charts/atlantis/values.yaml
```diff
@@ -172,6 +172,10 @@ serviceAccountSecrets: {}
 # Default values for atlantis (override as needed).
 ## -------------------------- ##
 
+# -- When set, ignores the port specified by the Service's targetPort attribute.
+# This is relevant when another container is injected into the same Pod.
+containerPort:
```
GMartinez-Sisti, line 177: This field should be of type integer alone and when it’s the default type value of zero it wouldn’t be used.

GMartinez-Sisti (commented):
Hi @jullianow, thank you for your contrubtion.

Please check the comment below and also:
- bump chart minor version due to new feature
- update the comment to reflect the new feature
  > -- (int) [optional] Define the port you would like atlantis to run on. Defaults to 4141.

charts/atlantis/values.yaml
```diff
@@ -172,13 +172,16 @@ serviceAccountSecrets: {}
 # Default values for atlantis (override as needed).
 ## -------------------------- ##
 
+# -- (int) [optional] Define the port you would like atlantis to run on. Defaults to 4141.
```
GMartinez-Sisti, line 175:
```suggestion
# -- (int) [optional] Define the port you would like atlantis to run on. When 0, the value of service.targetPort is used.
```

charts/atlantis/values.schema.json
```diff
@@ -267,6 +267,15 @@
         }
       }
     },
+    "containerPort": {
```
GMartinez-Sisti, line 270: Type needs to be integer alone with default as 0.

GMartinez-Sisti (commented):
Thanks for the update, just missing some details.

GMartinez-Sisti (approved):
Thanks @jullianow!

PR #446, opened 2024-12-22
Use stringData instead of data for secrets

afarid:
## what

This PR replaces the use of "data" to use "stringData" for the secrets

## why

This allows postrendering script to work as it can parse the secrets values and replace them, For example helmfile vals
https://github.com/helmfile/vals

## tests

- test run manually by comparing secrets values before and after the changed

## references

- (https://github.com/helmfile/vals)


GMartinez-Sisti (commented):
Thanks for the contribution @afarid. Please bump the patch version for the chart as this is more of a fix, than a feature.

Also from my understanding this is not a breaking change, if anyone thinks it is please do share the reasoning.

PR #455, opened 2025-02-11
Adapt labels to kubernetes recommended ones

jonitich:
## what

- Set Kubernetes recommended labels.

## why

- Align standards.

## tests

- I've testes it by rendering the helm chart.

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 appVersion: v0.47.1
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 6.15.0
+version: 6.16.0
 keywords:
   - terraform
 home: https://www.runatlantis.io
--- a/charts/atlantis/README.md
+++ b/charts/atlantis/README.md
@@ -21,6 +21,7 @@
   - [Install Helm Chart](#install-helm-chart)
 - [Required Configuration](#required-configuration)
 - [Additional manifests](#additional-manifests)
+- [Labels](#labels)
 - [Values](#values)
 - [Upgrading](#upgrading)
   - [From `4.*` to `5.*`](#from-4-to-5)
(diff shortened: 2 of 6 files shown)


GMartinez-Sisti (changes requested):
Hi @jonitich, thank you for the contribution, unfortunately users might be matching the current labels so we can't remove them.

I'll bring this to a discussion on the community so we can decide if it's worth a breaking change, a feature toggle or if it won't go through. I'll update the slack thread id here.

Update: slack thread -> https://cloud-native.slack.com/archives/C07T45G27EZ/p1739283078077209

jonitich:
Thanks for the quick response @GMartinez-Sisti, 
Br,
Joni.

jonitich:
hi @GMartinez-Sisti, I can see tthe  `Lint adn Test Charts` failing,  is there anything I can do about it?

GMartinez-Sisti:
> hi @GMartinez-Sisti, I can see tthe `Lint adn Test Charts` failing, is there anything I can do about it?

The error is:

> ✖︎ atlantis => (version: "5.14.0", path: "charts/atlantis") > chart version not ok. Needs a version bump! 

This version bump would be dependent on how this PR will be categorized. So no need to worry about this now.

jonitich:
hi @GMartinez-Sisti, it is done now, I've added other changes in order to keep all resources aligned when it comes to labels.
Thanks in advance.

GMartinez-Sisti:
Hi @jonitich, I think the best approach is to start by adding the new labels only, after some time we can drop the old ones in a non backwards compatible update. Please rebase and bump again.

jonitich:
Hi @GMartinez-Sisti, done — rebased onto latest `main` and reworked the approach as suggested:

- The Kubernetes-recommended labels (`app.kubernetes.io/name`, `app.kubernetes.io/instance`, `app.kubernetes.io/managed-by`, `app.kubernetes.io/version`) are now added alongside the existing legacy labels (`app`, `chart`, `release`, `heritage`), not replacing them.
- Selectors (StatefulSet, Service, PodMonitor) are left untouched for now, so this is a non-breaking, additive-only change.
- Chart version bumped to `6.16.0` (minor, since nothing breaking).
- Added a "Labels" section to the README documenting the change and noting the legacy labels will be dropped in a future major release.

Let me know if you'd like anything adjusted.

GMartinez-Sisti:
> Hi @GMartinez-Sisti, done — rebased onto latest `main` and reworked the approach as suggested:
> 
> * The Kubernetes-recommended labels (`app.kubernetes.io/name`, `app.kubernetes.io/instance`, `app.kubernetes.io/managed-by`, `app.kubernetes.io/version`) are now added alongside the existing legacy labels (`app`, `chart`, `release`, `heritage`), not replacing them.
> * Selectors (StatefulSet, Service, PodMonitor) are left untouched for now, so this is a non-breaking, additive-only change.
> * Chart version bumped to `6.16.0` (minor, since nothing breaking).
> * Added a "Labels" section to the README documenting the change and noting the legacy labels will be dropped in a future major release.
> 
> Let me know if you'd like anything adjusted.

Hi @jonitich, this is already going to recreate the statefulset pod, right? So I would be in favor of just adding it everywhere, and leave the current "legacy" ones behind a feature flag that will be enabled by default for now (e.g. `.Values.includeLegacyLabels`). This way we should break anything as the old labels will still be there, but we create a path forward to eventually remove them. WDYT?

Also please include the suggestion from https://github.com/runatlantis/helm-charts/pull/455#issuecomment-2666463716, I think this is a nice one.

PR #458, opened 2025-02-20
feat(annotations): add secret annotations

al-cheb:
## what

Allow adding annotations to secret resources.

## why

My use case is that I want to use vault-webhool to reflect secrets without define annotations for all resources when we use the extraAnnotations param.


charts/atlantis/README.md
```diff
@@ -183,6 +183,7 @@ extraManifests:
 | replicaCount | int | `1` | Replica count for Atlantis pods. |
 | repoConfig | string | `""` | Use Server Side Repo Config, ref: https://www.runatlantis.io/docs/server-side-repo-config.html. Check values.yaml for examples. |
 | resources | object | `{}` | Resources for Atlantis. Check values.yaml for examples. |
+| secretAnnotations | object | `{}` | These annotations will be added to secrets. Check values.yaml for examples. |
```
GMartinez-Sisti, line 186: What do you think of using `secret.Annotations` instead? I'm wondering if we will need more `secret` properties and then we just add them. E.g. `labels`.

GMartinez-Sisti (approved):

Issue #258, opened 2023-01-27
lifecycle hook to wait until all terraform processes are completed


nitrocode:
Lifecycle hooks are supported but it would be nice to add

- either a default lifecycle to prevent terraform from being killed in the middle of an apply or plan
    - otherwise we could have resources created and the information is not saved into the tfstate
    - otherwise we could have stale dynamodb locks
- or document the lifecycle hook to be optionally added

cc: @Balsir since you added the lifecycle feature to this repo, have you figured out a good `preStop` lifecycle ?

References
- https://github.com/runatlantis/helm-charts/pull/220
- https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/

nitrocode:
Yes, graceful termination should be implemented in the server. I was thinking this lifecycle method could be a nice workaround for now. :smile:

GenPage:
@nitrocode Should we close this issue and create one on the Atlantis repo?

nitrocode:
I think we can have both issues. One here for lifecycle method and one for graceful termination in the server.

mubarak-j:
currently, this helm chart supports pod lifecycle hooks: https://github.com/runatlantis/helm-charts/blob/4afd2a82b1fcc31026c4c392e6984895818b3e62/charts/atlantis/values.yaml#L489-L490

mubarak-j:
This is what I came up with, let me know if seems okay and I will create a PR.

```yaml
statefulSet:
  terminationGracePeriodSeconds: 180

lifecycle:
  preStop:
    exec:
      command:
        - /bin/bash
        - -c
        - while [ "$(pgrep -f terraform)" ]; do sleep 1; done
```

keviiin38:
I slightly adapted @mubarak-j's version in order to support Terraform, OpenTofu and Terragrunt

```yaml
terminationGracePeriodSeconds: 300

lifecycle:
  preStop:
    exec:
      command:
        - /bin/sh
        - -c
        - while pgrep -x "terraform|tofu|terragrunt" > /dev/null; do sleep 1; done
```

nitrocode:
If it has been tested, please put in a pr so everyone can take advantage. Great stuff @mubarak-j and @keviiin38 

Any additional thoughts on that too @Balsir @GenPage ?

keviiin38:
[...]

PR #468, opened 2025-04-20
docs(values.yaml): add preStop lifecyle example usage (#258)

keviiin38:
## what

- Implement the workaround proposed here : https://github.com/runatlantis/helm-charts/issues/258#issuecomment-2816793848 as a commented example

## why

- If any of Terraform, OpenTofu or Terragrunt commands are running, a StatefulSet upgrade should be delay until `terminationGracePeriodSeconds` is reached

## tests

- [x] I have tested my changes by adding a Helm test

## references

Relates #258


nitrocode (dismissed):

charts/atlantis/values.yaml
```diff
@@ -331,10 +331,6 @@ statefulSet:
   # -- Option to share process namespace with atlantis container.
   shareProcessNamespace: false
 
-# -- (int) Optionally customize the termination grace period in seconds.
-# @default -- default depends on the kubernetes version.
-terminationGracePeriodSeconds:
```
nitrocode, line 336: Do these docs need to be regenerated? Is this deprecated ?

keviiin38: No nothing changed here, I just moved it a little bit lower in the file, so it's just up to the lifecycle section

GMartinez-Sisti (commented):
@keviiin38 thanks for the contribution. Please bump the patch version of the chart and sign your commit. Details are in the failed jobs.

I’m not sure that if changing the order of the keys will require a docs update, after the above is sorted we’ll see what the linter says.

keviiin38:
> @keviiin38 thanks for the contribution. Please bump the patch version of the chart and sign your commit. Details are in the failed jobs.
> 
> I’m not sure that if changing the order of the keys will require a docs update, after the above is sorted we’ll see what the linter says.

@GMartinez-Sisti Done :+1:

(_`make docs` didn't generate changes locally_)

nitrocode:
Documentation requires a new version of the helm chart?

keviiin38:
> Documentation requires a new version of the helm chart?

@nitrocode I would rather not bump the version as this does not add a feature nor a fix, but as stated in https://github.com/runatlantis/helm-charts/pull/468#pullrequestreview-2785460526, the tests were failing...

GMartinez-Sisti:
> > Documentation requires a new version of the helm chart?
> 
> @nitrocode I would rather not bump the version as this does not add a feature nor a fix, but as stated in [#468 (review)](https://github.com/runatlantis/helm-charts/pull/468#pullrequestreview-2785460526), the tests were failing...

Every change on the chart requires a new version, even though is only documentation it is a change. This one is a small one but

GMartinez-Sisti:
@keviiin38 looks like now you're just missing the "signoff" in your commit. Details -> https://github.com/runatlantis/helm-charts/pull/468/checks?check_run_id=41030690365.

GMartinez-Sisti (approved):

nitrocode:
>Every change on the chart requires a new version, even though it's only documentation it is a change. This one is a small one but it could be huge so the process forces it.

Yes understood, but why ? It seems more like a consequence of the current implementation rather than the desired outcome. The reason why releases from sole documentation changes is not ideal is because every time a new version is released, people who subscribe to the release feed will get pinged. When I get pinged on a new release, I look at the release notes, and if I only see doc changes on a minor or patch release, I have to roll my eyes and ignore the release. Not only that, renovate/dependabot will create pull requests to bump the chart in every downstream implementation in argocd/flux which is very noisy.

It would be better to augment the release so the github actions do not run if

- ONLY markdown changes are made
- no new inputs
- no change to existing inputs
- no change to the logic itself

One way to do it would be to create a stripped down artifact, where comments are stripped from the chart itself. Then when the next change comes in, it can then check if it made a difference in the previous artifact, if so, cut a new change, if not, then no release.

Thoughts?

GMartinez-Sisti:
> Thoughts?

While I understand what you mean, this is a common practice across OSS repositories, and the docs update might even be useful for someone. This one particularly shows how to do something that you might’ve never thought about or were even having a problem with, and sometimes it is to signal that something important is going to happen or change.

The pipeline is currently pretty simple: something changes in the chart and we create a release. We can customize as much as we want but then we also have to maintain it. The suggested logic also means we need to validate that the chart is not being bumped when the change won’t create a release 😅.

If we are going to update the pipeline I would rather put some effort into making it smarter to avoid having to bump manually, either by requiring labels or the conventional commit format. Unfortunately I don’t have the time for it now, but I’m happy to help reviewing PRs!

Issue #421, opened 2024-09-10
Allow to change 'readOnly' parameter for 'aws-volume' volumeMount


rkubik-hostersi:
When using Atlantis in AWS EKS, the docker image has to be modified to include `aws-cli`. However, `aws` binary needs write access to the `~/.aws` directory to write some cache files when assuming roles.

Currently, this dir is mounted here: https://github.com/runatlantis/helm-charts/blob/a4ff46cf8deb67cef1f5427300824253a8b340b2/charts/atlantis/templates/statefulset.yaml#L566C1-L566C18 - it would be nice to make a variable out of this.

So far, when trying to call `aws` with assuming roles, I get following error:
```
[Errno 30] Read-only file system: '/home/atlantis/.aws/cli'
```

j0can:
Is this solved, i have the same issue trying to run atlantis on eks resources using a service account:
`[Errno 30] Read-only file system: '/home/atlantis/.aws/cli'`

PR #484, opened 2025-08-15
fix: unwritable .aws directory

### LINKED ISSUE #380: aws cli cache directory unwritable

gurchik:
## what

Mounts the AWS `credentials` and `config` secrets as individual files within a writable `.aws` directory, rather than mounting them in a read-only directory.

This PR has no effect if someone mounts an arbitrary AWS secret using `.Values.awsSecretName` because when doing that there is no way for the Chart to know the contents of the secret so it can't know which files to mount. So in this case there is no change of behavior.

## why

The AWS CLI cannot work properly when the `.aws` directory is read-only. This is a problem since this tool is often needed in Terraform projects. For example, a common way to configure the Helm Terraform provider is to pass in an `exec` argument which executes the AWS CLI to get Kubernetes credentials. Without this change, that would be impossible as the AWS CLI would error.

## tests

I tested my changes by deploying this to my cluster which was previously deploying version 5.18.1 of this Helm Chart.

Values:

```yaml
aws:
  config: |
    redacted
  credentials: |
    redacted
```

Before this change:

```
$ kubectl exec -n atlantis atlantis-0 -- aws sts get-caller-identity

[Errno 30] Read-only file system: '/home/atlantis/.aws/cli'
command terminated with exit code 255
```

After this change:

```
$ kubectl exec -n atlantis atlantis-0 -- aws sts get-caller-identity
{
    "UserId": "redacted",
    "Account": "redacted",
    "Arn": "redacted"
}
```

## references

closes https://github.com/runatlantis/helm-charts/issues/380
resolves dupe https://github.com/runatlantis/helm-charts/issues/421

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 appVersion: v0.35.1
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 5.18.1
+version: 5.18.2
 keywords:
   - terraform
 home: https://www.runatlantis.io
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 appVersion: v0.46.0
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 6.10.0
+version: 6.10.1
 keywords:
   - terraform
 home: https://www.runatlantis.io
--- a/charts/atlantis/templates/statefulset.yaml
+++ b/charts/atlantis/templates/statefulset.yaml
@@ -130,6 +130,10 @@ spec:
         secret:
           secretName: {{ .Values.netrcSecretName }}
       {{- end }}
+      {{- if or .Values.aws.credentials .Values.aws.config }}
+      - name: aws-dir
+        emptyDir: {}
+      {{- end }}
       {{- if or .Values.aws.credentials .Values.aws.config .Values.awsSecretName }}
       - name: aws-volume
         secret:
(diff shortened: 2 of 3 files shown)


GMartinez-Sisti:
Hi @gurchik, thank you for the PR. I'll try to take a look at it during the weekend.

gurchik:
@GMartinez-Sisti any update on this? :) I will bump the chart version once it's reviewed

GMartinez-Sisti:
> @GMartinez-Sisti the checksum fails because your CI workflow hardcodes expected SHA sums that don't match what the maintainer uploaded for the helm-unittest release [here](https://github.com/helm-unittest/helm-unittest/releases/tag/v1.0.3). I understand that the hardcoded SHA sums used to match until the helm-unittest maintainer replaced the release artifacts. I don't know why they replaced the artifacts, but I don't think their release workflow creates reproducible builds. Therefore, I find it unlikely the maintainer will be able to get the release to match the previous checksum now that it has been changed. If you are concerned about a potential supply chain attack then feel free to continue to investigate, if not then I would recommend updating your hardcoded checksums with the new values.

Right, we need confirmation from the maintainer that this is intended and have the checksum file updated or get a new release. This forced update goes against the best practices for released software, an artifact should never change, a new release should be published instead.

GMartinez-Sisti:
Hi @gurchik, https://github.com/runatlantis/helm-charts/pull/524 has unblock this. Please rebase so we can check the tests.

GMartinez-Sisti:
Hi @gurchik, please rebase and I’ll take a look.

gurchik:
@GMartinez-Sisti done

charts/atlantis/Chart.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: v1
 appVersion: v0.46.0
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 6.9.3
+version: 6.10.0
```
GMartinez-Sisti: Something got merged in the meantime:
```suggestion
version: 6.10.1
```

gurchik:
@GMartinez-Sisti Next week will be the one year anniversary of opening this PR. We keep looping between me rebasing it and someone approving a different more recent PR, causing me to need to rebase again. Can we please merge this soon?

GMartinez-Sisti (approved):
Thank you for your patience 🙏

PR #489, opened 2025-08-25
Add image.registry field

jcoelho93:
## what

This change adds the `image.registry` value to the helm-chart, and changes the Atlantis StatefulSet to use this new value when composing the Atlantis image name to pull.

This change is **backwards compatible**. No breaking change since the registry value defaults to ghcr.io/

## why

In cases where you have a multi-region/multi-environment deployment and use container image caches, you want to be able to specify different registries per region/environment. This way you can override only the image registry and keep the image name the same across environments.

Yes, this is _syntatic sugar_ more than anything but it's a pattern that's been used a lot.

## tests

- [x] ran `helm template chart/atlantis` and checked the resulting manifests
- [x] apply the helm chart on my local minikube and checked atlantis pull the correct image

## references

n/a

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 appVersion: v0.35.1
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 5.18.2
+version: 5.19.0
 keywords:
   - terraform
 home: https://www.runatlantis.io


GMartinez-Sisti (changes requested):
Hi @jcoelho93, thank you for the change, unfortunately this can be a breaking change for users that are already using another repository for the atlantis image.

E.g. if I'm currently using image.repository: myrepo/atlantis, this change would make it ghcr.io/myrepo/atlantis, unless I set image.registry: "".

For this reason, please document the change in the README file and bump major.

PR #515, opened 2025-11-26
feat: Add extra secret manifests support

tejisin:
# feat: Add `extraSecretManifests` for native SOPS & Secret management

## what

*   **New Feature**: Added `extra-secret-manifests.yaml` template to the chart.
*   **Schema Update**: Updated `values.schema.json` to support and validate the new `extraSecretManifests` field.
*   **Capability**: Enables users to define and deploy arbitrary Kubernetes `Secret` resources directly via the chart. This is specifically designed to unlock workflows using **SOPS** encrypted secrets adjacent to the deployment.
*   **Impact**: Zero impact on existing users; this is a purely additive feature.

## why

*   **First-Class SOPS Support**: This change allows users to manage secrets securely using [SOPS](https://github.com/getsops/sops) right alongside their chart values.
*   **Enhanced Security**:
    *   By using a dedicated `Secret` manifest, users can configure SOPS to only encrypt the `data` key.
    *   **Solves the `stringData` vulnerability**: Using `stringData` for secrets causes the plaintext value to be stored in the `last-applied-configuration` annotation (see [HackerOne Report](https://hackerone.com/reports/1102064)). This approach avoids that entirely by using standard `data` fields.
*   **Clean Separation of Concerns**:
    *   **Avoids Collisions**: Using the existing `extraManifests` for secrets forces encryption on everything in the file (including ConfigMaps), which is often undesirable.
    *   **Prevents Integrity Errors**: SOPS injects a file hash into the metadata. If ConfigMaps were mixed with Secrets and modified outside SOPS, the hash validation would fail. This dedicated template ensures SOPS integrity checks only apply where they matter.

## tests

*   [x] I have tested my changes by running `helm template` with various configurations:

```bash
helm template . \
  -f values.yaml \
  -f path/to/my-sops-secrets.yaml \
  > rendered.yaml
```

*   **Result**: The chart renders perfectly with `extraSecretManifests` provided.
*   **Regression**: Confirmed that the chart renders exactly as before when the new value is omitted.

## references

*   [Mozilla SOPS (Secrets OPerationS)](https://github.com/getsops/sops)
*   [Kubectl apply reveals secrets in last-applied-configuration](https://hackerone.com/reports/1102064)

---

### Example SOPS Configuration
For reviewers/users, a minimal `sops.yaml` to leverage this feature effectively:

```yaml
creation_rules:
  - encrypted_regex: '^(data)$'
    version: 3.11.0
```

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6
         with:
           fetch-depth: 0
 
@@ -25,9 +25,9 @@ jobs:
         uses: azure/setup-helm@b9e51907a09c216f16ebe8536097933489208112 # v4
         with:
           # renovate: datasource=github-tags depName=helm packageName=helm/helm
-          version: v4.0.0
+          version: v4.0.1
 
-      - uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v6
+      - uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548 # v6
         with:
           # renovate: datasource=docker depName=python
           python-version: '3.14'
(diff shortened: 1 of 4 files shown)


champs (approved):

tejisin:
> why do we need this and just use `extraManifests`? there is no added value for this

as mentioned under clean separation of concerns under Why, mixing non secret and secret resources under extraManifests would lead to integrity errors for the sops checksum if a value for another manifest was changed outside of the sops flow. Also the fact that configmaps and secrets use the same fields for storage (data/stringData) which would lead to configmaps getting encrypted as well if colocated.

PR #505, opened 2025-11-07
add artifact attestation

DrFaust92:
closes https://github.com/runatlantis/helm-charts/issues/409

tested locally from fork:

Screenshot 2025-11-07 at 10 37 59

PR #520, opened 2025-12-11
fix(helm-unittest): support helm 4

GMartinez-Sisti:
## what

- Caused by https://github.com/helm-unittest/helm-unittest/issues/777
- Related to:
  - https://github.com/runatlantis/helm-charts/pull/505#issuecomment-3617293420
  - https://github.com/runatlantis/helm-charts/pull/519

## why

[Helm-unittest](https://github.com/helm-unittest/helm-unittest) is not fully compatible with helm 4, the current fix is to disable the verification of the package contents on install but this is something we don't want to perform this for security reasons.

The current testing flow uses a makefile step that will only install helm-unittest when it's not available, as a temporary fix this PR installs a pinned version while validating the checksum of the file to ensure integrity.

With this new version the namespace test needed a fix.

## tests

Check CI.

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -3,8 +3,8 @@ name: Lint and Test Charts
 on:
   pull_request:
     paths:
-      - 'charts/atlantis/**'
-      - 'ct.yaml'
+      - "charts/atlantis/**"
+      - "ct.yaml"
 
 defaults:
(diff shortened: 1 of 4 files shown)


GMartinez-Sisti:
Github is flaky -> https://www.githubstatus.com/incidents/xntfc1fz5rfb will retry later.

Issue #485, opened 2025-08-19
Helm lint fails at /properties/extraAnnotations


Kozoft:
Hi. 
I'm getting this error from `helm lint`:

```
❯ helm lint .
==> Linting .
[ERROR] values.yaml: "file:///values.schema.json#" is not valid against metaschema: jsonschema validation failed with 'https://json-schema.org/draft/2019-09/schema#'
- at '': 'allOf' failed
  - at '/properties/extraAnnotations': 'allOf' failed
    - at '/properties/extraAnnotations/examples': got object, want array
[ERROR] templates/: values don't meet the specifications of the schema(s) in the following chart(s):
atlantis:
"file:///values.schema.json#" is not valid against metaschema: jsonschema validation failed with 'https://json-schema.org/draft/2019-09/schema#'
- at '': 'allOf' failed
  - at '/properties/extraAnnotations': 'allOf' failed
    - at '/properties/extraAnnotations/examples': got object, want array

Error: 1 chart(s) linted, 1 chart(s) failed
```

Looks like `extraAnnotations/examples` should be an array, not an object, to pass the validation against metachema.

GMartinez-Sisti:
@Kozoft Can you please share what you are setting on this key?

Kozoft:
> [@Kozoft](https://github.com/Kozoft) Can you please share what you are setting on the `extraAnnotations` key?

@GMartinez-Sisti Nothing, I don't have `extraAnnotations` in my custom values.
Tested with `[]`, `{}`, and `name: value` as `extraAnnotations` content, always getting the same validation error.

GMartinez-Sisti:
Thank you for the details, but I'm not able to reproduce this:

```sh
→ git show HEAD --summary
commit aab3a02421840c77c00cdda11007f0481c0f0cfb (HEAD -> main, tag: atlantis-5.18.2, origin/main, origin/HEAD, atlantis/main)
Author: Gaston Festari <cilindrox@gmail.com>
Date:   Tue Aug 19 10:27:09 2025 -0300

    fix(json-schema): update extraArgs definition (#483)

    The "examples" and "type" definition of the extraArgs property seems to
    have been swapped with the extraAnnotations definition.

    This was causing certain workflows that rely on the JSON-Schema
    correctness to fail.

    Bump chart version to v5.18.2

    Signed-off-by: Gaston Festari <cilindrox@gmail.com>
```
```sh
→ helm lint .
==> Linting .

[...]

PR #522, opened 2025-12-17
Update schema for additional properties in Ingress

macneib:
## what

- Fixed JSON schema validation error in values.schema.json for the Atlantis Helm chart

- Corrected the structure of the ingress property definition where labels, path, paths, pathType, host, hosts, and tls were incorrectly nested inside annotations

- These properties are now properly defined as siblings of annotations at the correct level within ingress.properties

## why

- The schema file had a structural error where multiple properties were nested inside the annotations object instead of being siblings

- This caused metaschema validation failures with the error: 'allOf' failed - at '/properties/extraArgs/examples': got object, want array

- The malformed schema prevented successful CI validation and Helm chart deployment

- Fixing the schema ensures proper validation of Helm values files and prevents deployment failures

tests

-  I have tested my changes by validating the corrected JSON schema structure

>   - Verified that annotations is properly closed after its additionalProperties definition
>   -  Confirmed that labels, path, paths, pathType, host, hosts, and tls are now siblings of annotations
>   -  CI validation should pass with the corrected schema

## references

JSON Schema Draft 2019-09 specification: [https://json-schema.org/draft/2019-09/schema](https://json-schema.org/draft/2019-09/schema)

Kubernetes Ingress documentation: [https://kubernetes.io/docs/concepts/services-networking/ingress/](https://kubernetes.io/docs/concepts/services-networking/ingress/)

### DIFF
--- a/charts/atlantis/values.schema.json
+++ b/charts/atlantis/values.schema.json
@@ -679,6 +679,7 @@
           "description": "Additional annotations to use for the Ingress.",
           "additionalProperties": {
             "type": "string"
+            }
           },
           "labels": {
             "type": "object",


GMartinez-Sisti (changes requested):
Hi @macneib, I don't think this change is required. The issue you mentioned was fixed on https://github.com/runatlantis/helm-charts/issues/485. Can you confirm please?

PR #538, opened 2026-02-19
Add service.labels

scottmelhop:
## what

Add service.labels to add additional labels to service as required

## why

Need to add specific labels to service for my gateway

## tests

Added unit test

### DIFF
--- a/charts/atlantis/values.yaml
+++ b/charts/atlantis/values.yaml
@@ -303,6 +303,7 @@ readinessProbe:
 
 service:
   type: NodePort
+  labels: {}
   annotations: {}
   port: 80
   portName: atlantis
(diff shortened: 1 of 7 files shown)


charts/atlantis/values.yaml
```diff
@@ -758,14 +759,14 @@ servicemonitor:
       # keys:
       #   username: USERNAME
       #   password: ATLANTIS_WEB_PASSWORD
-  # -- Optional metric relabelings to drop or modify metrics.
+      # -- Optional metric relabelings to drop or modify metrics.
```
nvanheuverzwijn: Feels like it shouldn't be indented ?

GMartinez-Sisti:
Hi @scottmelhop, thank you for the contribution. If this is still valid please rebase and I’ll take a look.

PR #553, opened 2026-05-28
fix: don't create an empty `<appname>-webhook` secret with defaults

### LINKED ISSUE #552: Empty secret `atlantis-webhook` created when no vendor is defined in values
We don't use any git repository vendor dict in our setup

```yaml
# (taken from default)
github: {}
githubApp: {}
gitea: {}
gitlab: {}
bitbucket: {}
azuredevops: {}
```

We still get a `atlantis-webhook` secret generated with `data=null`.

It doesn't affect, just noisy. I can issue a PR to address this.

ylmrx:
## what

Using default values, we get a `<appname>-webhook` secret, with data=null.

We can avoid creating it entirely in this case.

## why

Leaner deployment, avoiding empty resources whenever possible

## tests

- Changed the default secret-webhook test case as the document isn't created
- Ensure the document isn't created whenever `vcsSecretName` is def'd
- Set a repository so labels can be checked (since defaults don't create the resource anymore)

## references

closes #552

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 appVersion: v0.43.0
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 6.5.0
+version: 6.5.1
 keywords:
   - terraform
 home: https://www.runatlantis.io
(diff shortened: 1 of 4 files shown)


GMartinez-Sisti:
Thank you @ylmrx! Can you please bump the chart patch, since this is a small fix?

GMartinez-Sisti (approved):

PR #560, opened 2026-06-24
feat(atlantis): add service.extraPorts for additional service ports

DrFaust92:
## what

- Add an optional `service.extraPorts` value to the atlantis chart. Entries are appended to the Service's `ports` list verbatim (standard Service port objects).
- Renders nothing when the list is empty, so the default Service output is unchanged.
- Updates `values.yaml`, `values.schema.json`, README (helm-docs), and adds a unittest.

## why

- The chart currently renders exactly one Service port (`service.port`/`service.targetPort`) with no way to expose additional ports. Wrapping the chart to front Atlantis with an auth-proxy sidecar (e.g. oauth2-proxy on `:4180`) forces consumers to create a second, parallel Service just to expose the proxy port — the upstream-managed Service can't carry it.
- `extraPorts` lets a single Service expose both the Atlantis app port (for the ServiceMonitor `/metrics` scrape on `portName: atlantis`) and a sidecar port, consolidating to one Service.

## tests

- [x] `helm lint` passes
- [x] `helm unittest` passes (122 tests; added an `extraPorts` case to `tests/service_test.yaml`)
- [x] `helm template` renders the extra port and the empty default is byte-identical to before
- [x] README regenerated with helm-docs

## references

- Additive and backward-compatible; mirrors the common `extraPorts` pattern used by other charts.

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 appVersion: v0.44.0
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 6.6.0
+version: 6.7.0
 keywords:
   - terraform
 home: https://www.runatlantis.io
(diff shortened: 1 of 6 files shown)


GMartinez-Sisti (approved):
Thank you!

PR #563, opened 2026-06-26
feat(atlantis): allow restartPolicy on containers for native sidecars

DrFaust92:
## What

The bundled `charts/atlantis/values.schema.json` carries an older Kubernetes `io.k8s.api.core.v1.Container` definition with `additionalProperties: false` that predates the `restartPolicy` field added to `Container` in Kubernetes **1.28**.

Because both `initContainers` and `extraContainers` `$ref` that definition, declaring an init container as a **native sidecar** (an `initContainers` entry with `restartPolicy: Always`, stable since k8s 1.29) is rejected at render time:

```
at '/initContainers/0': additional properties 'restartPolicy' not allowed
```

The `statefulset.yaml` template already renders `initContainers`/`extraContainers` verbatim via `toYaml`, so the only thing blocking native sidecars is the schema.

## Change

- Add `restartPolicy` (string) to the `io.k8s.api.core.v1.Container` definition in `values.schema.json`. One edit covers both `initContainers` and `extraContainers` since they share the `$ref`.
- Add a `statefulset` unittest asserting an init container with `restartPolicy: Always` renders (and, since helm-unittest validates against the schema, that it now passes validation).
- Bump chart version `6.7.1` → `6.7.2`.

Additive only — default rendering is unchanged.

## Testing

- `helm lint charts/atlantis` ✅
- `helm unittest charts/atlantis` ✅ (123 tests)

## Use case

Running an auth/connectivity sidecar (e.g. oauth2-proxy or cloud-sql-proxy) as a native sidecar so it is Ready before the atlantis container starts and drains after it during shutdown.

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 appVersion: v0.45.0
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 6.9.0
+version: 6.9.1
 keywords:
   - terraform
 home: https://www.runatlantis.io
--- a/charts/atlantis/tests/statefulset_test.yaml
+++ b/charts/atlantis/tests/statefulset_test.yaml
@@ -258,6 +258,20 @@ tests:
       - equal:
           path: spec.template.spec.hostNetwork
           value: true
+  - it: native sidecar via initContainers restartPolicy
+    template: statefulset.yaml
+    set:
+      initContainers:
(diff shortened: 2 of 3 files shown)


GMartinez-Sisti:
Hi @DrFaust92. Thanks for the contribution, can you please rebase from main?

DrFaust92:
> Hi @DrFaust92. Thanks for the contribution, can you please rebase from main?

Thanks, Done

GMartinez-Sisti (approved):

Issue #459, opened 2025-02-25
Pod fails to start when using helm chart customPem value


RoFz:
Helm chart `customPem` value is set to:

```yaml
# -- Allows to override the /etc/ssl/certs/ca-certificates.cer with your custom one.
# You have to create a secret with the specified name.
customPem: gitlab-server-cert
```

The respective `gitlab-server-cert` secret is created as generic/Opaque in Atlantis' namespace.

```yaml
apiVersion: v1
data:
  gitlab.local.domain.crt: <base_64_cert_data>
kind: Secret
metadata:
  creationTimestamp: "2025-02-24T15:11:47Z"
  labels:
    kustomize.toolkit.fluxcd.io/name: apps
    kustomize.toolkit.fluxcd.io/namespace: flux-system
  name: gitlab-server-cert
  namespace: atlantis
  resourceVersion: "140336"
  uid: 6d8e52ad-9865-474e-8493-8474919579a6
type: Opaque
```

Atlantis pod fails to start with:

```
NAME         READY   STATUS             RESTARTS      AGE
atlantis-0   0/1     CrashLoopBackOff   4 (76s ago)   2m57s
```

`  Warning  Failed       20h (x4 over 20h)  kubelet            Error: failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/var/lib/kubelet/pods/7b61111e-3af1-4960-a271-0e9b878c171a/volume-subpaths/additional-trust-certs/atlantis/1" to rootfs at "/etc/ssl/certs/ca-certificates.crt": create mountpoint for /etc/ssl/certs/ca-certificates.crt mount: cannot create subdirectories in "/run/containerd/io.containerd.runtime.v2.task/k8s.io/atlantis/rootfs/etc/ssl/certs/ca-certificates.crt": not a directory: unknown`
 

Atlantis v0.33.0 (chart version 5.14.0) running on:
kind v0.26.0 go1.23.4 darwin/arm64 running on:
podman engine 5.3.2 go1.23.5 darwin/arm64

Related [PR](https://github.com/runatlantis/helm-charts/pull/163/commits/4b077dd372a7e1b361827242a0c7598d7c7a454a)?

GMartinez-Sisti:
Hi, looks the same as this https://github.com/runatlantis/helm-charts/pull/163#issuecomment-1218212055 ?

RoFz:
[...]

PR #568, opened 2026-06-30
feat(atlantis): add additionalCaCerts to append custom CAs to the trust bundle (non-root)

### LINKED ISSUE #567: Append custom CA certificates to the system trust bundle (non-root)
### Problem

When Atlantis runs behind a TLS-inspecting proxy/firewall (e.g. Azure Firewall application rules, Zscaler, Netskope) that re-signs egress traffic with a **private/internal CA**, every outbound TLS call fails with:

```
Error: initializing server: Get "https://.../api/v4/version": x509: certificate signed by unknown authority
```

This affects the `atlantis` server itself plus every subprocess it spawns — `terraform`, `git`, `curl`/`wget`, and Python `requests`-based tooling such as `checkov` and `az`.

The chart already offers `customPem`, but it **replaces** the entire `/etc/ssl/certs/ca-certificates.crt` via a `subPath` mount. That has two problems:

1. It drops every public CA baked into the image, breaking TLS to any host that is *not* re-signed by the private CA.
2. It is fragile — see #459 (pod fails to start) — and the original request #82 was only ever satisfied by this replace-style approach.

A correct solution must **append** the private CA(s) to the existing system bundle, and must work as the **default non-root user** (uid 100) — i.e. without `update-ca-certificates`, `runAsUser: 0`, or a `hostPath` volume.

### Proposed solution

Add an `additionalCaCerts` value:

- A small init container (defaulting to the Atlantis image, so the system CA set stays in sync) concatenates the image's `ca-certificates.crt` with the `*.crt` files from a referenced secret into a shared `emptyDir`.
- The Atlantis container is pointed at the merged bundle via the relevant trust env vars: `SSL_CERT_FILE` (Go: atlantis/terraform, OpenSSL, Python stdlib), `GIT_SSL_CAINFO` (git), `CURL_CA_BUNDLE` (curl/wget), `REQUESTS_CA_BUNDLE` (Python requests: checkov, az), with an `extraEnvVars` escape hatch for additional runtimes shipped in custom images.

This keeps public CAs intact, runs non-root, and requires no image changes. `customPem` stays for backward compatibility.

### Notes

- Revisits #82 and supersedes the replace-style limitation behind #459.
- I have a working, tested implementation (helm-unittest + lint) and will open a PR that closes this issue.

emanuelbesliu:
## what

- Add a new `additionalCaCerts` value to the Atlantis chart that **appends** one or more custom CA certificates to the system trust bundle, instead of **replacing** it (as the existing `customPem` does).
- An init container (defaulting to the Atlantis image) concatenates the image's `/etc/ssl/certs/ca-certificates.crt` with the `*.crt` files from a referenced secret into a shared `emptyDir`.
- The Atlantis container is pointed at the merged bundle via the relevant trust env vars — `SSL_CERT_FILE`, `GIT_SSL_CAINFO`, `CURL_CA_BUNDLE`, `REQUESTS_CA_BUNDLE` — with an `extraEnvVars` escape hatch for additional runtimes shipped in custom images.
- `customPem` is left untouched for backward compatibility.
- Adds `values.schema.json` entries, helm-unittest coverage, regenerated `README.md`, and bumps the chart `version` `6.9.2` → `6.10.0`.

## why

- When Atlantis runs behind a TLS-inspecting proxy/firewall (Azure Firewall application rules, Zscaler, Netskope, …) that re-signs egress traffic with a **private/internal CA**, every outbound TLS call fails with `x509: certificate signed by unknown authority`. This breaks the `atlantis` server and every subprocess it spawns: `terraform`, `git`, `curl`/`wget`, and Python `requests`-based tooling (`checkov`, `az`).
- The existing `customPem` **replaces** the whole `ca-certificates.crt` via a `subPath` mount, which (a) drops every public CA baked into the image — breaking TLS to any host that is *not* re-signed — and (b) is fragile (see #459, pod fails to start).
- `additionalCaCerts` keeps the public CA set intact (and in sync with the image), works as the **default non-root user** (uid 100) — no `update-ca-certificates`, no `runAsUser: 0`, no `hostPath` — and requires no image changes.

## tests

- [x] `helm lint ./charts/atlantis` passes with the feature both disabled (default) and enabled.
- [x] `helm unittest ./charts/atlantis` — full suite passes (127/127), including a new `tests/additional-ca-certs_test.yaml` asserting: nothing rendered when disabled; init container, secret volume, `emptyDir`, env vars and read-only mount rendered when enabled; custom init image override; `extraEnvVars` appended.
- [x] `values.schema.json` validation rejects malformed input (e.g. `additionalCaCerts.enabled=notabool`).
- [x] `README.md` regenerated with helm-docs (v1.14.2).
- [x] Verified end-to-end on a live cluster behind an Azure Firewall doing TLS inspection: reproducing the exact rendered init-container script + env vars as the non-root `atlantis` user, every tool (`atlantis server`, `terraform`, `git`, `curl`, `checkov`/Python `requests`, `az`) failed without the merged bundle and succeeded with it against a firewall-inspected host.

## references

- closes #567
- Revisits #82 (Support custom CA certs) and addresses the replace-style limitation behind #459 (pod fails to start with `customPem`).

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 appVersion: v0.45.0
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 6.9.2
+version: 6.10.0
 keywords:
   - terraform
 home: https://www.runatlantis.io
--- a/charts/atlantis/README.md
+++ b/charts/atlantis/README.md
@@ -241,6 +241,12 @@ Then point the browser-facing Ingress host at the `atlantis-ui` service port (`8
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
+| additionalCaCerts.enabled | bool | `false` | Enable appending custom CA certificate(s) to the system trust bundle. |
+| additionalCaCerts.envVars | list | `["SSL_CERT_FILE","GIT_SSL_CAINFO","CURL_CA_BUNDLE","REQUESTS_CA_BUNDLE"]` | Core env vars set on the Atlantis container, pointed at the merged bundle. Covers Go (atlantis, terraform), OpenSSL/C (curl, wget), git and Python requests-based tooling (e.g. checkov, az). |
+| additionalCaCerts.extraEnvVars | list | `[]` | Additional env vars to point at the merged bundle, for extra runtimes that ship in custom images (e.g. NODE_EXTRA_CA_CERTS, AWS_CA_BUNDLE, CLOUDSDK_CORE_CUSTOM_CA_CERTS_FILE). Harmless if the tool is absent. |
+| additionalCaCerts.image | string | `""` | Image used by the init container that builds the merged CA bundle. Defaults to the Atlantis image so the system CA set matches the app exactly and no extra image needs to be pulled/mirrored. |
+| additionalCaCerts.mountPath | string | `"/ca-certs/ca-certificates.crt"` | Path the merged CA bundle is written to (inside a shared emptyDir). |
+| additionalCaCerts.secretName | string | `""` | Name of a secret containing one or more `*.crt` PEM files to trust. The secret has to be created manually. Every key ending in `.crt` is appended. |
 | affinity | object | `{}` |  |
 | allowDraftPRs | bool | `false` | Enables atlantis to run on a draft Pull Requests. |
 | allowForkPRs | bool | `false` | Enables atlantis to run on a fork Pull Requests. |
--- a/charts/atlantis/templates/statefulset.yaml
+++ b/charts/atlantis/templates/statefulset.yaml
@@ -157,6 +157,13 @@ spec:
         secret:
           secretName: {{ .Values.customPem }}
       {{- end }}
+      {{- if .Values.additionalCaCerts.enabled }}
+      - name: additional-ca-certs
+        secret:
+          secretName: {{ .Values.additionalCaCerts.secretName }}
+      - name: ca-certs-bundle
+        emptyDir: {}
+      {{- end }}
       {{- if .Values.extraVolumes }}
       {{- toYaml .Values.extraVolumes | nindent 6 }}
       {{- end }}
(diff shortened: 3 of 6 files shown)


GMartinez-Sisti:
HI @emanuelbesliu, I'm wondering if this a valid addition to the helm chart. I understand your requirement, but at the same time you have the required configurations to add this via config (`extraVolumes`, `initContainers`, `etc`).

Also my experience tells me that a business that has a TLS inspecting proxy will also likely require the container images to be pulled from a secure and controlled private registry, so they would have the infrastructure to build their own custom atlantis image with the certificate baked in.

emanuelbesliu:
Hi @GMartinez-Sisti, thanks for the thoughtful review — you're right, and I want to concede the main point up front.

Everything this PR does can already be assembled from existing extension points (`extraVolumes` for the secret + `emptyDir`, `initContainers` for the bundle builder, `extraVolumeMounts` for the merged bundle, and `environmentRaw` for the trust env vars). It's a convenience wrapper, not a missing capability, and I don't want to grow the chart's surface area for something that's already achievable.

Your point about private registries is fair too, though I'd gently note that baking the CA into a custom image couples cert rotation to image rebuilds, and plenty of orgs mirror the upstream image through a pull-through cache without maintaining their own Dockerfile. So a secret-driven bundle is still useful — but that's a "nice to have," not a justification for new chart values.

The part I think *does* carry real value isn't the six values — it's the pattern itself. The existing `customPem` **replaces** the whole `ca-certificates.crt` via a `subPath` mount, which drops every public CA baked into the image and is fragile (#459). The non-obvious, safe approach is to **append** custom CAs to the system bundle, keep it in sync with the image, and do it all as the default non-root user. That's exactly what people get wrong when they hit this behind a TLS-inspecting proxy.

So rather than new chart logic, would you be open to me converting this into a **documentation recipe** in the chart README — a copy-pasteable `extraVolumes` / `initContainers` / `extraVolumeMounts` / `environmentRaw` example showing the append-style, non-root pattern? That transfers the knowledge without adding anything for you to maintain long-term. I'm happy to drop all the template/schema changes and reduce this PR to docs (I can keep or drop the helm-unittest, your call).

If you'd prefer, I can also just close this and open a focused docs-only PR. Let me know which you'd like.

GMartinez-Sisti:
Hi @emanuelbesliu, thank you for the reply. I would add this to the docs instead. Feel free to reuse this PR and adjust the code and description.

PR #570, opened 2026-07-02
feat: Add default automerge method config

ronaldmiranda:
## what

 - add the `automergeMethod` value

## references

https://github.com/runatlantis/atlantis/pull/6573


GMartinez-Sisti:
Hi @ronaldmiranda, thank you for the contribution but we want to keep the chart simple to use. Atlantis has a lot of options and adding them all to the helm chart would make it very complex and harder to manage ([Server Configuration Docs](https://www.runatlantis.io/docs/server-configuration.html)).

To add extra environment variables you can use any of the following:
- `.Values.loadEnvFromConfigMaps`
- `.Values.loadEnvFromSecrets`
- `.Values.environment`
- `.Values.environmentRaw`
- `.Values.environmentSecrets`

ronaldmiranda:
@GMartinez-Sisti yes i know, but i just wanted to keep the same pattern.

PR #576, opened 2026-07-21
feat(atlantis): support named service ports in ingress backends

sakkiii:
## Description

Adds support for referencing a Kubernetes Service backend by **named port** (in addition to the existing numeric port) in the Atlantis chart Ingress and webhook Ingress templates.

Previously, ingress backend ports were always rendered as `port.number`, which meant users had to switch to the custom `paths` structure just to use a named service port. This change lets the simple/default ingress path stay simple while still allowing a string (named) port.

## Changes

- Add `atlantis.ingressServicePort` helper in `_helpers.tpl` that renders either `name: <string>` or `number: <int>` depending on the value type.
- Use the helper in `templates/ingress.yaml` and `templates/webhook-ingress.yaml` for the default and custom-path backends.
- Allow `string` (named port) in addition to `integer`/`null` for the `port` field in `values.schema.json`.

## Behavior

- Numeric port values render as `port.number` (unchanged, backwards compatible).
- String port values render as `port.name`.

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 appVersion: v0.46.0
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 6.10.0
+version: 6.10.1
 keywords:
   - terraform
 home: https://www.runatlantis.io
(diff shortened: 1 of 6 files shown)


GMartinez-Sisti (commented):
Hi @sakkiii, can you add a simple test to ensure this logic works as expected? A small test with a named port confirming that the result matches should be enough, Also please rebase and bump again as something eles got merged. Thank you

sakkiii-paypay:
@GMartinez-Sisti done

GMartinez-Sisti (approved):

PR #579, opened 2026-07-23
feat(atlantis): add optional PodDisruptionBudget

mxibrahim:
## what

- Adds a new optional `PodDisruptionBudget` template for the Atlantis StatefulSet, disabled by default.
- Configurable via a new `podDisruptionBudget` values block: `enabled`, `minAvailable`, `maxUnavailable` (mutually exclusive), plus `labels` and `annotations`.
- When `maxUnavailable` is set it takes precedence; otherwise `minAvailable` is used (default `1`). Both accept integers or percentages.
- Includes the JSON schema entry, `helm-unittest` tests, and updated docs/README. Bumps chart version `6.9.3` → `6.10.0`.

## why

- A PodDisruptionBudget is a common, expected resource for production-grade charts and was the main gap here — it lets operators keep Atlantis available during voluntary disruptions such as node drains and cluster upgrades.
- Kept opt-in and off by default so there is no behavior change for existing users. Note is included in `values.yaml` that, because Atlantis is stateful and usually runs a single replica, enabling a PDB with `minAvailable: 1` will block voluntary evictions of that pod.

## tests

- [x] Ran `helm unittest ./charts/atlantis` — 130/130 tests pass across 15 suites, including the new `tests/pdb_test.yaml` (7 cases: disabled-by-default, defaults, `minAvailable` override, `maxUnavailable` precedence, percentage values, labels/annotations, `commonLabels`).
- [x] Ran `helm lint ./charts/atlantis` — passes, confirming the new values validate against `values.schema.json`.
- [x] Verified rendered output for both `minAvailable` and `maxUnavailable` modes via `helm template`.

## references

N/A

### DIFF
--- a/charts/atlantis/Chart.yaml
+++ b/charts/atlantis/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 appVersion: v0.46.0
 description: A Helm chart for Atlantis https://www.runatlantis.io
 name: atlantis
-version: 6.9.3
+version: 6.10.0
 keywords:
   - terraform
 home: https://www.runatlantis.io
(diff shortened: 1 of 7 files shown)


GMartinez-Sisti (approved):
While I don’t like PDBs that block kubernetes maintenances and operations, I understand the need in this particular application. Thank you!

